PR #74, opened 2026-02-19
refactor: activity-driven timer & boundary event registration

nightBaker:
## Summary

- **Move timer & boundary event registration into Activity.ExecuteAsync** — each activity now registers its own boundary events (timers + messages) during execution, consistent with how intermediate catch events already work. Replaces centralized post-execution batch scans in `WorkflowInstance.ExecuteWorkflow()`.
- **Introduce `TimerCallbackGrain`** — stateless `IRemindable` grain (`IGrainWithGuidCompoundKey`) that manages Orleans reminders and calls back to `WorkflowInstance.HandleTimerFired`. Removes `IRemindable` from `WorkflowInstance`. Key uses `hostActivityInstanceId` for uniqueness per activity instance, supporting future loop scenarios.
- **Fix boundary cleanup asymmetry** — `HandleBoundaryMessageFired` now also cancels boundary timer callbacks for the interrupted activity (was only cleaning up boundary messages before).
- **Reliable retry via periodic reminders** — `TimerCallbackGrain` calls back to `WorkflowInstance` before unregistering the reminder. If the callback fails, the periodic reminder retries. `HandleTimerFired` is idempotent via stale-timer guards.
- **Enforce single-execution invariant** — `ActivityInstanceState.Execute()`, `Complete()`, and `Fail()` now throw `InvalidOperationException` on invalid state transitions (double execution, double completion, etc.). Documented as a design constraint in `CLAUDE.md`.

## Commits

1. `refactor: move timer & boundary event registration into Activity.ExecuteAsync` — core refactoring
2. `refactor: use activityInstanceId in TimerCallbackGrain key` — unique key per host activity instance
3. `refactor: rename RegisterTimerReminder params for clarity` — `hostActivityInstanceId` + `timerActivityId`
4. `docs: add single-execution activity invariant to CLAUDE.md`
5. `docs: fix activity invariant wording — instances vs definitions`
6. `feat: add state transition guards to ActivityInstanceState`

## Test plan

- [x] `dotnet build` — zero errors
- [x] `dotnet test` — all 279 tests pass (72 Domain + 57 Infrastructure + 90 Persistence + 60 Application)
- [x] `BoundaryTimerEventTests` — uses `HandleTimerFired` instead of `IRemindable.ReceiveReminder`
- [x] `TimerIntermediateCatchEventTests` — new `HandleTimerFired_ShouldCompleteWorkflow` test
- [x] `MessageBoundaryEventTests` — boundary message flow unchanged
- [x] Removed `Update_ClearsErrorState` persistence test (tested now-invalid Fail→Execute transition)

### DIFF
--- a/src/Fleans/Fleans.Application.Tests/MessageBoundaryEventTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/MessageBoundaryEventTests.cs
@@ -117,4 +117,57 @@ public class MessageBoundaryEventTests : WorkflowTestBase
         var delivered = await correlationGrain.DeliverMessage("order-789", new ExpandoObject());
         Assert.IsFalse(delivered, "Subscription should have been cleaned up");
     }
+
+    [TestMethod]
+    public async Task BoundaryMessage_DirectCallAfterCompletion_ShouldBeSilentlyIgnored()
+    {
+        // Arrange — Start → Task(+BoundaryMessage) → End, BoundaryMessage → MsgEnd
+        var start = new StartEvent("start");
+        var task = new TaskActivity("task1");
+        var msgDef = new MessageDefinition("msg1", "cancelOrder", "orderId");
+        var boundaryMsg = new MessageBoundaryEvent("bmsg1", "task1", "msg1");
+        var end = new EndEvent("end");
+        var msgEnd = new EndEvent("msgEnd");
+
+        var workflow = new WorkflowDefinition
+        {
+            WorkflowId = "boundary-msg-stale",
+            Activities = [start, task, boundaryMsg, end, msgEnd],
+            SequenceFlows =
+            [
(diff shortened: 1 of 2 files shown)


src/Fleans/Fleans.Application/Grains/TimerCallbackGrain.cs
```diff
@@ -0,0 +1,87 @@
...
+        if (reminderName != ReminderName)
+            return;
+
+        var workflowInstanceId = this.GetPrimaryKey(out var activityId);
+        LogReminderFired(workflowInstanceId, activityId!);
+
+        // Unregister the reminder first
+        try
+        {
+            var reminder = await this.GetReminder(ReminderName);
```
nightBaker: what about periodic reminders @claude, should we unregister always ?

PR #113, opened 2026-02-27
Feature/multi instance activity

nightBaker:


nightBaker:
## Code Review Fixes (1a61f10)

Addressed all findings from code review:

**Critical:**
- **C1:** Fixed `FailActivityWithBoundaryCheck` entry lookup — now uses `activityInstanceId` to resolve the correct MI iteration entry instead of `GetFirstActive(activityId)` which could pick the wrong iteration
- **C2:** Added `ParallelCardinality_WhenIterationFails_ShouldSetErrorState` test — verifies error code (500) and message on iteration failure

**Important:**
- **I1:** Added constructor validation on `MultiInstanceActivity` — requires either `LoopCardinality` or `InputCollection`, rejects negative cardinality
- **I2:** Changed `MultiInstanceActivity` base class from `Activity` to `BoundarableActivity` — boundary events now register on MI hosts
- **I3:** Added `LogMultiInstanceIterationCompleted` (EventId 1043) log message for iteration completions
- **I4:** Collection test now verifies output aggregation values (`results` variable with 3 items)
- **I5:** Defensive null check on variable state before output aggregation read

**Minor:**
- **M1:** Test assertions use exact count (`== 4`) instead of `>= 3`
- **M3:** Combined two `ExpandoObject` merges into single `MergeState` call

All 380 tests pass, 0 build errors.

PR #128, opened 2026-03-11
Feature/workflow execution aggregate

nightBaker:
## Summary

Major architectural refactoring that extracts BPMN workflow orchestration logic from the `WorkflowInstance` grain into a `WorkflowExecution` domain aggregate, following DDD and Clean Architecture principles.

### Key Changes

- **WorkflowExecution aggregate** (Domain layer): Pure domain logic with Emit/Apply event sourcing pattern. 17 domain event types, 12 infrastructure effect types. Zero infrastructure dependencies.
- **Thin grain coordinator**: `WorkflowInstance` grain delegates all domain logic to the aggregate, performing only infrastructure effects (timers, message/signal subscriptions, child workflows).
- **ActivityInstanceEntry enrichment**: Absorbs execution state from the deleted `ActivityInstance` grain, eliminating per-activity grain overhead.
- **Infrastructure effects as return values**: Aggregate never calls infrastructure directly — returns `IReadOnlyList<IInfrastructureEffect>` for the grain to perform.
- **Timer cycle deadlock fix**: Integrates PR #132's approach — `HandleTimerFired` returns `TimeSpan?` so `TimerCallbackGrain` handles cycle re-registration locally (no reentrant grain call).
- **Intent-collecting adapter**: `ActivityExecutionContextAdapter` collects intent flags during activity execution; the grain reads them afterward and routes all state changes through Emit/Apply.

### Deleted Code

- `ActivityInstance` grain and `ActivityInstanceState`
- `BoundaryEventHandler` service and `IBoundaryEventStateAccessor`
- `EfCoreActivityInstanceGrainStorage`
- Consolidated grain from 6 to 3 partial files

### Test Coverage

- 299 domain-level unit tests (aggregate logic)
- 153 application-level integration tests (grain coordinator)
- All 452 tests passing

### CLAUDE.md Compliance

- `[LoggerMessage]` source generators for all state mutations
- Domain state methods self-contained (atomic flag + timestamp updates)
- Partial class layout: `WorkflowInstance.cs` (entry points), `WorkflowInstance.Infrastructure.cs` (effects/loop), `WorkflowInstance.Logging.cs` (log declarations)

### DIFF
--- a/src/Fleans/Fleans.Application/Grains/WorkflowInstance.Infrastructure.cs
+++ b/src/Fleans/Fleans.Application/Grains/WorkflowInstance.Infrastructure.cs
@@ -303,7 +303,7 @@ public partial class WorkflowInstance
                     LogStateCompleted();
                     break;
                 case ActivitySpawned spawned:
-                    LogStateAddEntries(1);
+                    LogActivitySpawned(spawned.ActivityInstanceId, spawned.ActivityId, spawned.ActivityType);
                     break;
                 case ActivityCompleted completed:
                     LogStateCompleteEntries(1);
--- a/src/Fleans/Fleans.Application/Grains/WorkflowInstance.Logging.cs
+++ b/src/Fleans/Fleans.Application/Grains/WorkflowInstance.Logging.cs
@@ -205,6 +205,10 @@ public partial class WorkflowInstance
         Message = "Timer cycle re-registration deferred to TimerCallbackGrain for activity {TimerActivityId}, dueTime={DueTime}")]
     private partial void LogTimerCycleReRegistrationDeferred(string timerActivityId, TimeSpan dueTime);
 
+    [LoggerMessage(EventId = 3020, Level = LogLevel.Debug,
+        Message = "Activity spawned: ActivityInstanceId={ActivityInstanceId}, ActivityId={ActivityId}, ActivityType={ActivityType}")]
+    private partial void LogActivitySpawned(Guid activityInstanceId, string activityId, string activityType);
+
     [LoggerMessage(EventId = 3017, Level = LogLevel.Information, Message = "Workflow execution started")]
     private partial void LogExecutionStarted();
 


nightBaker:
## Follow-up Review Findings — PR #128

Deeper analysis uncovered several issues related to the Emit/Apply pattern consistency:

### 🔴 Blocking

**1. `MarkExecutionStarted()` bypasses Emit/Apply, violating aggregate's own pattern**

`WorkflowExecution.MarkExecutionStarted()` directly calls `_state.Start()` without going through `Emit()`/`Apply()`. This means the state mutation is invisible to `GetUncommittedEvents()` and won't be logged by `LogAndClearEvents()`. This violates the aggregate's core contract that every state mutation goes through Emit/Apply, and violates the CLAUDE.md rule: "Log all workflow instance state changes. No silent state mutations."

**2. `ResolveTransitions` mutates entry directly, bypassing Emit/Apply**

In `ResolveTransitions()`, `existingEntry.ResetExecuting()` is called directly on the entry object instead of going through an Emit/Apply event. This is a silent state mutation that won't be captured in uncommitted events or logged. Should be a domain event (e.g., `ActivityExecutionReset`).

**3. `SetChildWorkflowInstanceId` call in `ProcessStartChildWorkflow` bypasses Emit/Apply**

`entry.SetChildWorkflowInstanceId(childInstanceId)` directly mutates state without an event. Per the aggregate pattern, this should be a domain event (e.g., `ChildWorkflowLinked`).

### 🟡 Important

**4. Potential infinite loop in `RunExecutionLoop`**

The execution loop has no iteration limit. While `CompleteFinishedSubProcessScopes` has a 100-iteration guard, the outer `RunExecutionLoop` does not. If an activity produces commands that spawn pending activities that never complete, this loop runs forever. Consider adding a safety bound.

**5. Domain event records with mutable reference types (`ExpandoObject`) break event immutability**

Records like `ActivityCompleted(... ExpandoObject Variables)` hold mutable reference types. `ExpandoObject` is mutable by design. After event creation, external code could modify the variables, corrupting the event. Consider deep-cloning in `Emit()` or using immutable dictionaries.

**6. `FailActivity` parent notification check may be incorrect**

After `Emit(new ActivityFailed(...))`, the failed entry is already marked completed. So `_state.GetActiveActivities().Any()` may return empty even if other entries were genuinely still active before the failure. The logic assumes this is the last active entry, but this needs validation for cases where boundary error handlers spawn new activities.

**7. Timer cycle regression (Test 15C)**

The manual test results document a potential regression where timer cycles don't fire correctly. The `HandleTimerFired` -> return `TimeSpan?` -> `TimerCallbackGrain` re-register approach is architecturally sound, but this should be investigated before merge.

### 🟢 Minor

**8. `internal void MarkCompleted()` on `ActivityInstanceEntry` appears to be dead code** — the `Complete()` method fully encapsulates the completion transition. If nothing calls `MarkCompleted()`, it should be removed.

**9. Duplicate error classification logic** — Error mapping appears in both `ActivityInstanceEntry.Fail(Exception)` and `WorkflowExecution.FailActivity()`. Consider extracting to a single `ErrorClassifier`.

nightBaker (commented):
## PR #128 Review: Feature/Workflow Execution Aggregate

This is a significant and well-structured architectural refactoring. The extraction of domain logic into a `WorkflowExecution` aggregate with an Emit/Apply event-sourcing pattern is a strong move toward DDD and Clean Architecture. The elimination of per-activity `ActivityInstance` grains is a meaningful performance improvement. Overall, the design is sound.

Below are findings organized by severity.

---

### Blocking Issues

**1. `NotifyParentFailedEffect` stores a raw `Exception` object**
`InfrastructureEffects.cs` defines:
```
public record NotifyParentFailedEffect(
    Guid ParentInstanceId, string ParentActivityId,
    Exception Exception) : IInfrastructureEffect;
```
While effects are not serialized across grain boundaries today (they are computed and consumed in-memory within the same grain activation), storing a raw `Exception` in a record is fragile. `Exception` objects are not guaranteed to be serializable in all .NET environments, and if effects are ever persisted or logged as structured data, this will break. More importantly, the aggregate (a pure domain object) is accepting and passing through `Exception` directly in `FailActivity` and `OnChildWorkflowFailed`, which leaks infrastructure concerns into the domain layer.

**Recommendation:** Extract error code and message in the grain coordinator before calling the aggregate. The aggregate's `FailActivity` could accept `(int errorCode, string errorMessage)` instead of `Exception`, and `NotifyParentFailedEffect` could store the same. This would also make the aggregate fully testable without infrastructure types.

**2. Timer cycle regression (Test 15C)**
The manual test results document a new bug: timer cycle (`R3/PT5S`) does not fire. The `HandleTimerFired` return-value approach for cycle re-registration is clever for avoiding the deadlock, but the test results suggest it may not be working correctly. This is a functional regression introduced by this PR.

**Recommendation:** Investigate and fix before merge, or document as a known issue with a follow-up ticket.

---

### Important Issues

**3. `SnapshotExpandoObject` is shallow-copy only**
In `WorkflowExecution.Emit()`, `SnapshotExpandoObject` copies only the top-level key-value pairs. If any value is itself mutable (e.g., a nested `ExpandoObject`, `List<object>`, or `Dictionary`), the snapshot shares the same reference. This means domain events are not truly immutable if workflows use nested variable structures.

**Recommendation:** Consider a deep-copy implementation, or document this as a known limitation with a TODO.

**4. `ActivityExecutionContextAdapter` has `[GenerateSerializer]` but holds `List<IDomainEvent>`**
The adapter is marked `[GenerateSerializer]` with `[Id(4)]` on `_publishedEvents`, which is a `List<IDomainEvent>`. Since `IDomainEvent` implementations (e.g., `ExecuteScriptEvent`) are records without `[GenerateSerializer]`, this will fail if Orleans tries to serialize the adapter across grain boundaries. The adapter is returned from `GetActiveActivities()`/`GetCompletedActivities()` which are grain interface methods, so serialization will happen.

**Recommendation:** Either add `[GenerateSerializer]` to all `IDomainEvent` implementations that could appear in `_publishedEvents`, or clear `_publishedEvents` before returning adapters from state facade methods (since published events are only relevant during execution, not when querying state).

**5. Domain events lack `[GenerateSerializer]`**
The 17 domain event records in `WorkflowDomainEvents.cs` (e.g., `WorkflowStarted`, `ActivitySpawned`, etc.) do not have `[GenerateSerializer]`. Per CLAUDE.md rules, anything crossing grain boundaries needs this. While these events currently stay in-memory within the aggregate, the `PublishDomainEventEffect` wraps an `IDomainEvent` and `EvaluateConditionEvent`/`ExecuteScriptEvent` (which do cross boundaries via `IEventPublisher`) should have it. Verify all event types that flow through `PublishDomainEvent` have proper serialization attributes.

**6. `WorkflowExecution` class is 1,290 lines**
While the aggregate correctly consolidates domain logic, a single 1,290-line class is large. The class handles activity lifecycle, boundary events, scope completion, condition sequences, child workflows, token management, and variable scoping. Consider splitting into partial files by concern (similar to how the grain was previously organized) to improve navigability.

**7. Multiple `WriteStateAsync` calls per operation**
In `RunExecutionLoop`, `_state.WriteStateAsync()` is called at the end of each loop iteration. Additionally, `PerformMessageSubscribe` and `PerformSignalSubscribe` call `_state.WriteStateAsync()` before external subscription calls. The outer grain methods (e.g., `CompleteActivity`) also call `_state.WriteStateAsync()` at the end. This could result in 3+ state writes per single operation. While necessary for crash recovery, it may impact performance.

**Recommendation:** Document the intentional multi-write pattern with a comment explaining the crash-recovery rationale.

---

### Minor Issues

**8. `PendingActivity` and `CompletedActivityTransitions` records defined in aggregate file**
These records (`PendingActivity`, `CompletedActivityTransitions`) are defined at the top of `WorkflowExecution.cs` outside the class. They are public API of the domain layer and would be more discoverable in their own file or in a shared types file.

**9. Typo in existing file name preserved**
`WorfklowEvaluateConditionEventHandler.cs` (note "Worfklow" instead of "Workflow") is modified but the typo is not fixed. Minor, but worth a rename while touching the file.

**10. `EnsureExecution` pattern is slightly inconsistent**
`SetWorkflow` creates `_execution` directly, while all other public methods use `EnsureExecution()`. The `SetWorkflow` approach is correct (since it creates the definition first), but adding a brief comment explaining why `EnsureExecution` is not used there would help future maintainers.

**11. `CLAUDE.md` partial class layout update is accurate**
The CLAUDE.md update correctly reflects the new 3-file layout. Good.

---

### What's Done Well

- **Emit/Apply event sourcing pattern** is clean and well-implemented. All state mutations go through `Emit()` -> `Apply()`, ensuring consistency.
- **Infrastructure effects as return values** is an excellent pattern. The aggregate never calls infrastructure directly, making it fully unit-testable.
- **Intent-collecting adapter** (`ActivityExecutionContextAdapter`) is a clever bridge that avoids coupling activities to the aggregate's event system.
- **Timer deadlock fix** via return-value approach is architecturally sound.
- **Test coverage** is comprehensive: 299 domain tests + 153 integration tests. The domain tests thoroughly cover the aggregate's Emit/Apply paths.
- **Manual test plan** with 22+ scenarios and honest documentation of regressions and known bugs.
- **Clean deletion** of `ActivityInstance` grain, `BoundaryEventHandler`, and `IBoundaryEventStateAccessor` -- significant complexity removal.
- **`LogAndClearEvents`** properly logs every domain event type, satisfying the "no silent state mutations" rule.
- **Boundary event handling** is unified into a single `HandleBoundaryEventFired` method, reducing duplication.

---

### Verdict

The timer cycle regression (Test 15C) and the `Exception` in domain layer concern are worth addressing. The serialization issues (points 4-5) could cause runtime failures when querying active activities. I recommend fixing or explicitly tracking these before merge.

PR #132, opened 2026-03-11
fix: resolve timer cycle deadlock on non-interrupting boundary events

### LINKED ISSUE #130: Bug: Timer cycle (R3/PT5S) not firing on non-interrupting boundary event
## Description

Timer cycle (R3/PT5S) on a non-interrupting boundary event does not fire. Found during manual testing (Test 15C) on 2026-03-11.

## Steps to Reproduce
1. Deploy workflow `ni-timer-cycle-test` with `R3/PT5S` cycle boundary on a task
2. 2. Start the workflow
3. 3. Wait 20+ seconds
## Expected
Timer fires 3 times at 5s intervals, `sendReminder` executes each time.

## Actual
After 20+ seconds, no cycle timer fires occurred. Only `start` completed, `longTask` remains active.

## Investigation
Check `TimerCallbackGrain` cycle handling and whether the aggregate refactor (PR #128) affected timer re-registration.

## Source
Manual test report: `docs/plans/2026-03-11-manual-test-results-aggregate-refactor`
**Potential new bug** — may be caused by the aggregate refactor.

nightBaker:
## Summary

Fixes #130 — Timer cycle (R3/PT5S) on non-interrupting boundary events caused an Orleans grain deadlock.

**Root cause:** `BoundaryEventHandler.HandleBoundaryTimerFiredAsync` called `callbackGrain.Activate()` on the same `TimerCallbackGrain` currently executing `ReceiveReminder`. Since Orleans grains are non-reentrant by default, both awaits blocked forever (deadlock).

**Secondary issue:** `ReceiveReminder` unconditionally unregistered the reminder after `HandleTimerFired`, which would race with any re-registration performed during the boundary handler.

**Fix (Approach A from issue analysis):** Return `TimeSpan?` from `HandleTimerFired` instead of calling back to the originating grain. `TimerCallbackGrain` handles cycle re-registration locally after unregistering the old reminder.

### Files changed
- `IBoundaryEventHandler.cs` / `BoundaryEventHandler.cs` — Return `Task<TimeSpan?>` from `HandleBoundaryTimerFiredAsync`, removed direct `Activate()` call
- `IWorkflowInstanceGrain.cs` / `WorkflowInstance.EventHandling.cs` — Propagate `TimeSpan?` through the call chain
- `TimerCallbackGrain.cs` — Capture cycle re-registration result, re-register locally after unregister
- `BoundaryEventHandlerTests.cs` — Updated cycle timer tests to verify return value pattern

### Test results
- Build: 0 errors, 0 warnings
- Tests: 493 passed, 0 failed

## Test plan
- [x] `dotnet build` — zero errors
- [x] `dotnet test` — all 493 tests pass
- [ ] Manual test 15C (timer cycle R3/PT5S) — verify cycle fires 3 times at 5s intervals

### DIFF
--- a/src/Fleans/Fleans.Application.Tests/BoundaryEventHandlerTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/BoundaryEventHandlerTests.cs
@@ -371,7 +371,7 @@ public class BoundaryEventHandlerTests
     // --- Non-interrupting cycle timer re-registration tests ---
 
     [TestMethod]
-    public async Task HandleBoundaryTimerFired_NonInterruptingCycle_ShouldReRegisterTimer()
+    public async Task HandleBoundaryTimerFired_NonInterruptingCycle_ShouldReturnNextCycleDueTime()
     {
         // Arrange
         var hostInstanceId = Guid.NewGuid();
@@ -390,9 +390,6 @@ public class BoundaryEventHandlerTests
         var boundaryGrain = Substitute.For<IActivityInstanceGrain>();
         _grainFactory.GetGrain<IActivityInstanceGrain>(Arg.Is<Guid>(id => id != hostInstanceId)).Returns(boundaryGrain);
 
-        var timerCallbackGrain = Substitute.For<ITimerCallbackGrain>();
-        _grainFactory.GetGrain<ITimerCallbackGrain>(_state.Id, $"{hostInstanceId}:bt1").Returns(timerCallbackGrain);
-
         var definition = Substitute.For<IWorkflowDefinition>();
         definition.Activities.Returns(new List<Activity> { boundaryTimer });
         definition.SequenceFlows.Returns(new List<SequenceFlow>());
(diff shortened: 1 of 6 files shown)


nightBaker (commented):
## PR Review: fix: resolve timer cycle deadlock on non-interrupting boundary events

Good fix for the Orleans grain deadlock. The approach of returning `TimeSpan?` instead of calling back into the originating grain is clean and correct — it eliminates the re-entrant call that caused the deadlock and also fixes the reminder unregister-race as a side effect.

### Findings

#### 🟢 Minor

1. **No unit tests for `TimerCallbackGrain.ReceiveReminder` re-registration path.** The PR correctly tests that `BoundaryEventHandler` returns the right `TimeSpan?` values, but the new logic in `TimerCallbackGrain.ReceiveReminder` (lines 73-78 — the `if (cycleReRegistration.HasValue)` block that calls `RegisterOrUpdateReminder`) has no direct test coverage. This is the code that actually prevents the deadlock, so a grain-level test confirming `RegisterOrUpdateReminder` is called after a non-null return would strengthen confidence. Not blocking since the integration tests (493 passing) cover the end-to-end path, but worth adding later.

2. **`WorkflowInstance.EventHandling.cs` layout vs CLAUDE.md.** CLAUDE.md says `HandleTimerFired` (a public entry point) belongs in `WorkflowInstance.cs`, but it lives in `WorkflowInstance.EventHandling.cs`. This is a pre-existing layout issue (not introduced by this PR) — just noting it for future cleanup.

### What looks good

- **Deadlock fix is correct.** The return-value pattern cleanly avoids the non-reentrant grain callback. `TimerCallbackGrain` unregisters the old reminder first, then re-registers locally — correct ordering.
- **Logging compliance.** New `LogCycleReRegistering` uses `[LoggerMessage]` source generator (EventId 10005) as required. The state mutation (re-registration) is logged.
- **Test updates are appropriate.** The two `BoundaryEventHandlerTests` were updated to verify the return value pattern instead of mocking `ITimerCallbackGrain.Activate()`. The mock setup for `timerCallbackGrain` was correctly removed.
- **Interface change is clean.** `IWorkflowInstanceGrain.HandleTimerFired` signature change from `Task` to `Task<TimeSpan?>` is properly propagated through all layers: interface → grain → event handling → boundary handler → service interface.
- **Branch naming** follows `fix/` convention.
- **No domain state encapsulation violations.** The change stays in the application/grain layer and doesn't bypass domain methods.

LGTM — no blocking issues found. The minor items above are suggestions for future improvement.

PR #206, opened 2026-04-01
Implement pagination and filters for Workflows page (Sieve-based)

### LINKED ISSUE #174: pagination and filters for workflows page filter as in #139
#139

nightBaker:
## Summary

Closes #174

Reworks the Workflows page to use server-side pagination with Sieve-based filtering/sorting, replacing the previous client-side approach that loaded all process definitions into memory.

### Key changes:
- **New `ProcessDefinitionGroup` DTO** — groups all versions under a process key
- **New `GetProcessDefinitionGroups(PageRequest)` method** — hybrid Sieve filter + custom group-level sort/pagination
- **Sieve handles filtering** (`ProcessDefinitionKey@=`, `IsActive==`) via existing `ApplicationSieveProcessor` mappings
- **Custom group-level ordering** — explicit `OrderBy` on grouped keys (not relying on `GROUP BY` preserving order)
- **Workflows.razor** uses `GridItemsProvider` + `FluentPaginator` pattern (matching `ProcessInstances.razor`)
- **300ms debounced search** prevents excessive SignalR round-trips on keystroke
- **7 new persistence tests** covering pagination, search, filter, sort, empty results, and group completeness

### Design decisions:
- Uses standard `PageRequest` with Sieve instead of custom `ProcessDefinitionPageRequest` (per design feedback)
- `IsActive` filter applies to individual rows (any-version semantics, consistent with Sieve)
- Sort supports `ProcessDefinitionKey` (default) and `-DeployedAt`
- `ItemsProvider` callback pattern with `HierarchicalGridItem` expand/collapse preserved

### Test results:
- 7 new tests pass
- All 742 existing tests pass unchanged
- Build succeeds with 0 errors

## Test plan
- [ ] Verify workflows page loads with pagination (20 groups per page)
- [ ] Verify search by process key works with debounce
- [ ] Verify sort by deployed date works via column header
- [ ] Verify expand/collapse of version history still works
- [ ] Verify Start, Edit, Instances, Enable/Disable buttons work
- [ ] Verify pagination navigates correctly (next/prev/last page)


nightBaker (commented):
## Code Review: Implement pagination and filters for Workflows page

**Verdict: 🔄 Changes Suggested** — Good overall structure but two issues should be addressed before merge.

### What looks good
- Two-step query pattern (paginate keys → fetch versions) is sound for group-based pagination
- Correctly uses `IWorkflowQueryService` directly from Blazor (compliant with CLAUDE.md admin UI rule)
- `FluentButton` uses `IconStart` correctly (CLAUDE.md compliant)
- 7 new persistence tests cover the major scenarios
- `ProcessDefinitionPageRequest.Normalize()` correctly clamps inputs

---

### 🔴 Blocking — must fix

**B1. IsActive filter uses unfiltered `db.ProcessDefinitions` instead of the search-filtered `keysQuery`**
(`WorkflowQueryService.cs`)

The `latestVersionPerKey` subquery is built from the raw `db.ProcessDefinitions` table rather than from `keysQuery` (which already has the search filter applied). While this happens to produce correct results for current queries (because the final intersection catches it), it generates unnecessarily expensive SQL and is a correctness risk for future filter additions. The `latestVersionPerKey` should be derived from `keysQuery`, not `db.ProcessDefinitions`.

**B2. Search fires a server round-trip on every keystroke**
(`Workflows.razor`)

The `FluentTextField` uses `@bind-Value:after="OnSearchChanged"`, which triggers `LoadWorkflows()` on every keystroke. For Blazor Server, each character generates a SignalR round-trip + DB query, causing flickering, potential race conditions, and unnecessary server load.

**Recommended fixes:**
- Add debounce logic (e.g., 300ms `Timer`-based delay before `LoadWorkflows`)
- Or use `@onchange` (fires on blur/Enter) instead of `@bind-Value:after`

---

### 🟡 Important — should fix

1. **`ProcessDefinitionGroup` and `ProcessDefinitionPageRequest` missing `[GenerateSerializer]`** — CLAUDE.md: "`[GenerateSerializer]` on anything crossing grain boundaries." While these DTOs don't cross grain boundaries today, the convention should be followed for consistency with `ProcessDefinitionSummary`.

2. **`SortBy` parameter unused in UI** — The `ProcessDefinitionPageRequest` exposes `SortBy` and the query service handles it, but the Razor page never lets users change sort order. Either expose the UI or defer the server-side sort logic to a follow-up PR.

3. **No loading indicator during page transitions** — The spinner only shows on initial load. When navigating pages or changing filters, `isLoading` is set but no visual feedback appears. Users won't know the page is refreshing.

---

### 🟢 Minor

4. **Case-sensitive search** — `Contains(search)` is case-sensitive in SQLite (unlike SQL Server). The old client-side filter used `OrdinalIgnoreCase`. Consider `EF.Functions.Like` or `.ToLower()`.

5. **No test for combined search + IsActive filter** — Tests cover these independently but not together. A combined test would help catch the B1 interaction bug.

6. **Pagination buttons lack accessibility attributes** — Consider adding `AriaLabel` or `Title` for screen readers.

7. **`totalPages` uses local `PageSize` constant instead of `result.PageSize`** — Should use the server-returned value for consistency.

---

Please address B1 and B2 before merge. The Important items should ideally be addressed in this PR but could be deferred to a follow-up if scoped separately.

🤖 Reviewed by Claude Code (scheduled code review)

nightBaker:
## Code Review Fixes Applied

Addressed all findings from the code review:

### Blocking
- **B1 (Timer disposal):** Implemented `IDisposable` on `Workflows.razor` — `searchDebounceTimer` is now disposed when the component is torn down, preventing memory/callback leaks in Blazor Server circuits.

### Important
- **Y1 (Unfiltered Step 5):** Added clarifying comment explaining this is intentional — when a filter narrows which keys appear, all versions within matching keys are still shown for management purposes.
- **Y2 (Unnecessary StateHasChanged):** Removed `StateHasChanged()` calls inside the `ItemsProvider` callback (both in try and catch blocks). `FluentDataGrid` already handles re-rendering around provider calls.

### Minor
- **M1 (Missing `[GenerateSerializer]`):** Added `[GenerateSerializer]` and `[property: Id(N)]` attributes to `ProcessDefinitionGroup`, consistent with `ProcessDefinitionSummary`.
- **M2 (Count before ordering):** Moved `CountAsync()` before the sort step to avoid unnecessary `ORDER BY` in the count query.
- **M3 (Sort fallback comment):** Added comment explaining the default key sort fallback for unmapped columns in `BuildSortString`.

### Verification
- Build: 0 errors
- Tests: 742/742 passing (0 failures)

🤖 Fixed by Claude Code (scheduled developer pipeline)

PR #216, opened 2026-04-03
Fix unbounded variable state accumulation in sub-process scopes

### LINKED ISSUE #165: Scalability: Unbounded variable state accumulation without cleanup
## Problem
`WorkflowInstanceState.VariableStates` grows without bound as variable scopes are created for parallel forks, sub-processes, and multi-instance iterations via `AddCloneOfVariableState` (lines 128-145). Scopes are never removed, only superseded. Eagerly loaded on every state read, causing memory and serialization overhead proportional to the total number of scopes ever created.

## Affected Files
- `src/Fleans/Fleans.Domain/States/WorkflowInstanceState.cs:18, 128-145`
- `src/Fleans/Fleans.Persistence/EfCoreWorkflowInstanceGrainStorage.cs:24, 156-168`

## Recommendation
Track scope lifecycle and mark completed scopes as superseded. Lazy-load only active scopes on state read. Archive completed scopes to a separate table. Alternatively, implement garbage collection for unused variable states after dependent activities complete.

---
*Category: State Management & Persistence*
*Filed by scalability audit on 2026-03-15*

nightBaker:
Closes #165

## Summary

Sub-process child variable scopes were never removed after completion or cancellation, causing `VariableStates` to grow without bound — especially problematic for loops containing sub-processes.

### Changes

- **Sub-process completion path** (`CompleteFinishedSubProcessScopes`): Collect orphaned child scope IDs and return them for deferred removal. The grain removes them after all transitions are resolved in `RunExecutionLoop`, preventing a crash in nested sub-process scenarios where inner completion makes outer completion happen in the same call.
- **Interrupting boundary events** (`HandleBoundaryEventFired`): Emit `VariableScopesRemoved` immediately for cancelled sub-process child scopes, since the boundary event uses the host's parent scope.
- **Error boundary events**: No cleanup added — the boundary event is spawned with the failed activity's `VariablesId` (the child scope), so the scope must persist.
- Added `RemoveVariableScopes` method on `WorkflowExecution` aggregate for deferred scope removal.
- Updated `CompleteFinishedSubProcessScopes` return type to include `OrphanedScopeIds`.

### Key design decision

Scope removal must be **deferred** past transition resolution in the nested sub-process case. When both inner and outer sub-processes complete in the same `CompleteFinishedSubProcessScopes` call, the outer sub-process's child scope (A) is still needed by activities spawned from inner sub-process completion transitions. Removing A immediately caused `Variable state not found` errors.

### Tests

- 2 new integration tests verifying child scope cleanup after completion and interrupting boundary cancellation
- All 737 existing tests pass unchanged

## Test plan

- [ ] `SubProcessCompletion_CleansUpChildVariableScope` — verifies only root scope remains after sub-process completes
- [ ] `SubProcessCancellation_CleansUpChildVariableScope` — verifies only root scope remains after interrupting boundary cancels sub-process
- [ ] All existing sub-process, boundary event, and multi-instance tests pass
- [ ] Nested sub-process test (`SubProcess_NestedSubProcess_ShouldCompleteThrough`) passes

### DIFF
--- a/src/Fleans/Fleans.Application/Grains/WorkflowInstance.Infrastructure.cs
+++ b/src/Fleans/Fleans.Application/Grains/WorkflowInstance.Infrastructure.cs
@@ -38,8 +38,13 @@ public partial class WorkflowInstance
         }
 
         // Handle subprocess/multi-instance scope completions that may have occurred.
-        // Orphaned scope IDs are discarded here — they will be cleaned up by RunExecutionLoop
-        // which accumulates and removes them after all pending activities have completed.
+        // Orphaned scope IDs are intentionally discarded here (not in the main RunExecutionLoop):
+        // ResolveExternalCompletions runs *before* RunExecutionLoop on every external entry point
+        // (CompleteActivity, FailActivity, message/signal delivery, etc.). The subsequent
+        // RunExecutionLoop call will re-invoke HandleScopeCompletions on every iteration and
+        // accumulate any still-orphaned scope IDs into its own list, which is then drained via
+        // RemoveVariableScopes after the loop exits. Removing scopes here would risk pulling
+        // them out from under transitions that the loop is about to compute.
         await HandleScopeCompletions(definition);
     }
 
@@ -120,8 +125,10 @@ public partial class WorkflowInstance
 
         // Remove orphaned sub-process child variable scopes after the execution loop
         // completes. Scopes must persist during the loop because activities spawned by
-        // sub-process completion transitions still reference them.
-        _execution!.RemoveVariableScopes(allOrphanedScopeIds);
+        // sub-process completion transitions still reference them. The VariableScopesRemoved
+        // event raised by the aggregate is logged via LogEvent → LogVariableScopesRemoved.
+        if (allOrphanedScopeIds.Count > 0)
+            _execution!.RemoveVariableScopes(allOrphanedScopeIds);
     }
 
     /// <summary>
--- a/src/Fleans/Fleans.Domain/Aggregates/WorkflowExecution.cs
+++ b/src/Fleans/Fleans.Domain/Aggregates/WorkflowExecution.cs
@@ -1009,9 +1009,12 @@ public class WorkflowExecution
             effects.AddRange(CancelScopeChildren(result.HostActivityInstanceId));
 
             // Clean up child variable scope of the interrupted sub-process
+            // Filter to scopes that have a parent (i.e. true child scopes), rather than
+            // "not equal to host scope" — this is safer against entries that may reference
+            // an unrelated scope. Matches the convention used by sub-process completion cleanup.
             var interruptedChildScopes = _state.GetEntriesInScope(hostEntry.ActivityInstanceId)
                 .Select(e => e.VariablesId)
-                .Where(vid => vid != hostEntry.VariablesId)
+                .Where(vid => _state.VariableStates.Any(vs => vs.Id == vid && vs.ParentVariablesId.HasValue))
                 .Distinct()
                 .ToList();
             if (interruptedChildScopes.Count > 0)


nightBaker (commented):
## Code Review — PR #216 (re-review after new commit 7e617e9)

Second commit adds sub-process variable propagation (child → parent merge on completion). Clean implementation that builds naturally on the deferred scope cleanup from the first commit.

### 🟢 Highlights

- **Variable merge before completion** is the right sequencing — `VariablesMerged` emitted before `ActivityCompleted` ensures parent scope has child data before downstream activities reference it.
- **`ParentVariablesId` lookup** for finding the child scope (line ~596) is correct — it avoids dependency on scope entries that may have been removed by internal fork-join resolution.
- **Deferred scope removal** continues to work correctly — orphaned scope IDs accumulate across `HandleScopeCompletions` calls and are cleaned up after `RunExecutionLoop` exits.
- **Boundary event path** removes scopes immediately via `VariableScopesRemoved` — correct because boundary spawns activities using the host's parent scope, not the child scope.
- **Test coverage** is thorough: scope cleanup on completion, scope cleanup on cancellation, variable propagation, and internal parallel gateway propagation.
- **CLAUDE.md compliance**: Domain logic in `WorkflowExecution`, state mutation via `Emit`, existing `[LoggerMessage]` declarations cover the emitted events, tests follow AAA pattern.

### 🟡 Important

- **Test name mismatch**: `SubProcess_VariableShadowing_ShouldNotAffectParent` now asserts that the parent **is** affected (color = "blue" from child merge). The test name contradicts its assertion. Should be renamed to something like `SubProcess_VariableShadowing_ShouldMergeToParent` to match the new semantics.

### 🟢 Minor

- The `RemoveVariableScopes` guard (`scopeIds.Count > 0`) is good defensive coding. The corresponding guard at the call site in `RunExecutionLoop` (checking `allOrphanedScopeIds` before calling) would be a minor optimization but is not necessary since the guard is already in the domain method.
- Consider documenting the sub-process variable merge behavior in the design doc or README, since this is a semantic change from "isolated child scopes" to "merge-on-completion". Not blocking.

No blocking issues found. CI green, no conflicts. Ready to merge.

🤖 Reviewed by Claude Code

nightBaker:
## Review Feedback Addressed

Pushed commit `630c94c` addressing the remaining review items.

### 🟡 Important — Fixed

**Test name mismatch (`SubProcess_VariableShadowing_ShouldNotAffectParent`)**

Renamed to `SubProcess_VariableShadowing_ShouldMergeToParent` to match the actual assertion semantics — the test verifies that child scope variables *are* merged into the parent (color = "blue"), so the old name contradicted the behavior being tested.

---

### Answering reviewer question: "Do we remove orphanedScopeIds only after merging it to the main scope?"

**Yes, exactly.** The sequence in `CompleteFinishedSubProcessScopes` is:

1. **Merge first** — `VariablesMerged` event is emitted (line 633) copying child scope variables into the parent scope
2. **Collect for deferred removal** — the child scope ID is added to `allOrphanedScopeIds` (line 636), but **not removed yet**
3. **After `RunExecutionLoop` exits** — `RemoveVariableScopes(allOrphanedScopeIds)` is called (line 121 in Infrastructure.cs), which emits `VariableScopesRemoved`

This ordering guarantees that:
- Child scope data is merged into the parent *before* the scope is removed
- Activities spawned by sub-process completion transitions can still reference child scope variables during the execution loop
- Scope cleanup happens in a single batch after all transitions are resolved

All 13 SubProcess tests pass, including the renamed test.

nightBaker:
## Merge Conflicts Resolved + Reviewer Question Answered

### Merge conflicts with main resolved (commit 997d8b1)

Three files had conflicts after main's recent refactors:

1. **`WorkflowExecution.cs`** (CompleteFinishedSubProcessScopes + HandleBoundaryEventFired)
   - Combined main's defensive multi-child-scope check (`Count > 1` throws) with this PR's deferred orphaned-scope cleanup pattern (`allOrphanedScopeIds`).
   - Boundary event handler: kept main's `BoundaryEventService.HandleFired` refactor and re-applied the interrupting child-scope cleanup inside the new structure.

2. **`WorkflowExecutionScopeCompletionTests.cs`** — adopted main's `ScriptTask` convention while preserving the PR's full 3-tuple assertions (effects, completedHostIds, orphanedScopeIds) and child→parent merge semantics.

3. **`SubProcessTests.cs`** — kept BOTH the PR's new `SubProcessCompletion_CleansUpChildVariableScope` / `SubProcessCancellation_CleansUpChildVariableScope` tests AND main's `SubProcess_VariablePropagation_*`, `SubProcess_WithInternalParallelGateway_*`, and `SubProcess_Failure_ShouldNotMergeVariablesIntoParent` tests — they cover different semantics and don't overlap.

### Verification

```
Domain.Tests:        344 passed
Application.Tests:   207 passed
Infrastructure.Tests: 98 passed
Persistence.Tests:   116 passed
Mcp.Tests:             4 passed
Total:               769 passed, 0 failed
```

Build clean (0 errors).

### Reviewer question — answered

> Do we remove orphanedScopeIds only after merging it to the main scope?

**Yes — by design.** The flow is:

1. `CompleteFinishedSubProcessScopes()` walks active subprocess hosts. For each host whose children are fully completed, it:
   - Emits `VariablesMerged(parentVarId, childScope.Variables)` — merges child values into the parent scope.
   - Adds `childScope.Id` to `allOrphanedScopeIds` (deferred — does **not** emit `VariableScopesRemoved` here).
   - Emits `ActivityCompleted` for the host.

2. The method returns the orphaned IDs alongside effects + completed host IDs.

3. `RunExecutionLoop` (in `WorkflowInstance.Infrastructure.cs`) accumulates orphaned IDs across iterations and only calls `execution.RemoveVariableScopes(allOrphanedScopeIds)` **after the loop exits** — i.e., after all merges and all transitions for nested subprocess completions have been resolved.

The deferral is necessary because nested subprocess transitions resolved later in the same loop iteration may still reference the child scope (the interpreter walks parent chains). Removing it inline would crash those lookups.

So the invariant is: **merge first → defer removal until the loop is fully drained → then remove.** Variables are guaranteed to be in the parent scope before the child scope disappears.

🤖 Pushed by the review-developer scheduled task.

PR #235, opened 2026-04-05
Clean up orphaned child variable scopes on subprocess cancellation (#165)

### LINKED ISSUE #165: Scalability: Unbounded variable state accumulation without cleanup

nightBaker:
## Summary

Closes #165

- Added `RemoveChildVariableScopes()` helper method to `WorkflowExecution` that identifies and removes orphaned child variable scopes for entries in a given scope
- Called after `CancelScopeChildren` for **interrupting boundary events** (timer/message/signal) where the boundary handler uses the host's variable scope, making the child scope orphaned
- **Error boundaries are intentionally excluded** — the error handler inherits the failed entry's `VariablesId` (the child scope), so removing it would break variable resolution
- Added two tests:
  - `SubProcessCancellation_CleansUpChildVariableScope` — verifies child scope is removed when an interrupting boundary timer cancels a subprocess
  - `ErrorBoundary_OnSubProcess_PreservesChildVariableScope` — verifies child scope is preserved when an error boundary fires (handler still references it)

### Key design decision

The original plan suggested adding cleanup directly in `CancelScopeChildren()`. During implementation, I discovered this would break error boundaries: the error handler is spawned with `VariablesId: entry.VariablesId` (the child scope), so removing it causes a `Variable state not found` crash when the handler completes. The fix was extracted into a separate `RemoveChildVariableScopes()` method called only at safe call sites.

### Completion path verification (PR #218)

PR #218 (merged) handles the subprocess **completion** path — it merges child scope variables into the parent scope. The child scope is intentionally not removed on completion due to a timing issue with the do-while cascading loop (documented in a TODO comment).

## Test plan

- [x] All 337 domain tests pass
- [x] All 205 application (integration) tests pass
- [x] All 109 persistence tests pass
- [x] New tests verify both cleanup and preservation scenarios

### DIFF
--- a/docs/plans/2026-04-04-guaranteed-event-delivery-design.md
+++ b/docs/plans/2026-04-04-guaranteed-event-delivery-design.md
@@ -1,0 +1,120 @@
+# Guaranteed Event Delivery — Design
+
+**Issue:** [#160](https://github.com/nightBaker/fleans/issues/160)
+**Date:** 2026-04-04
+
+## Problem
+
+`WorkflowEventsPublisher` publishes domain events (`EvaluateConditionEvent`, `ExecuteScriptEvent`) to Orleans in-memory streams. If a handler crashes or the silo restarts mid-processing, these events are lost and the workflow stalls. No retry, no dead-letter queue, no replay.
+
+Note: aggregate state events are already durable via JournaledGrain + EfCoreEventStore. Only the two external processing events are at risk.
+
+## Approach: Pluggable Persistent Stream Providers
+
+Replace the hardcoded `WithMemoryStreaming("StreamProvider")` with a configuration-driven stream provider selection. Use existing Orleans stream adapter packages (EventHubs, SQS, community RabbitMQ/Kafka) — no custom adapters.
+
+## Design
+
+### 1. Stream Provider Pluggability
+
+A config key selects the provider:
+
+```json
+{
+  "Fleans": {
+    "Streaming": {
+      "Provider": "memory"
+    }
+  }
+}
+```
+
+Extension method in `Fleans.ServiceDefaults`:
+
+```csharp
+public static ISiloBuilder AddFleanStreaming(this ISiloBuilder builder, IConfiguration config)
+{
+    var provider = config.GetValue<string>("Fleans:Streaming:Provider") ?? "memory";
+    return provider switch
+    {
+        "memory" => builder.AddMemoryStreams("StreamProvider"),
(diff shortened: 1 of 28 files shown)


nightBaker (commented):
Reviewed against CLAUDE.md and the diff.

## 🔴 Blocking
- **Duplicate of PR #216.** PR #216 also closes #165 and adds the same cancellation cleanup inline in `HandleBoundaryEventFired`. The two PRs will conflict on `WorkflowExecution.cs`. Only one should be merged. PR #216 is broader (handles both completion and cancellation paths), so I'd recommend either:
  1. Closing this PR in favor of #216, or
  2. Rebasing this PR to only contribute the `ErrorBoundary_OnSubProcess_PreservesChildVariableScope` regression test, which #216 lacks and is genuinely valuable.

## 🟢 Positive
- The `RemoveChildVariableScopes` helper is cleaner than #216's inline block — it correctly filters via `ParentVariablesId.HasValue` rather than by "not equal to host scope," which is more defensive.
- The error-boundary preservation test is a great regression guard for the subtle interaction between error handlers and child scopes. Worth keeping in some form.
- Clear PR description explaining the design decision and the difference vs. error boundaries.

## 🟡 Important
- No completion-path cleanup. If #216 is closed in favor of this one, the sub-process *completion* path (the more common case) would still leak scopes. This PR is incomplete on its own for #165.
- New helper `RemoveChildVariableScopes` mutates state but has no `[LoggerMessage]` log call. Per CLAUDE.md: "Log all workflow instance state changes."

Requesting changes to resolve the overlap with #216 before merge.

PR #330, opened 2026-04-21
Add CSS-only motion and hover effects to docs website

### LINKED ISSUE #260: Add CSS-only motion and hover details
Staggered page-load reveals on hero, scroll-triggered accents, surprising hover states on nav and cards. Pure CSS (Starlight is not React).

Skill phase: Motion & Interaction.
Parent: #254

nightBaker:
## Summary

Closes #260.

- **Hero stagger reveals**: `fade-up-clip` on title/tagline/actions with 60-380ms stagger delays, `clip-reveal-x` on the logo image
- **Scroll-driven accents**: left-border draw on `.card` and underline draw on `h2::after` using `animation-timeline: view()` behind `@supports` guard (static fallback for unsupported browsers)
- **Hover / focus states**: sidebar lime `>` caret + text shift, `.site-title` underline wipe, `.card` lift + hard lime shadow, primary hero action underline wipe, copy-button blinking lime dot
- **Accessibility**: all motion gated on `prefers-reduced-motion: no-preference`, hover effects in `@media (hover: hover)`, `:focus-visible` mirrors always active, `cursor-blink` frozen under reduced-motion
- **DESIGN.md**: appended Motion section documenting principles, three moments, keyframes table, and accessibility approach

No JavaScript. No new dependencies. Website-only change (2 files).

## Test plan

- [ ] `npm run build` in `website/` passes (verified locally)
- [ ] Dark theme: hero elements stagger in on page load
- [ ] Light theme: same hero animation plays correctly
- [ ] Hover sidebar links: lime `>` caret appears, text shifts 4px
- [ ] Hover `.site-title`: accent-2 underline wipes in from left
- [ ] Hover `.card`: border color changes, lifts -1px/-1px, lime shadow appears
- [ ] Hover primary hero action: accent-2 underline wipes in
- [ ] Tab through interactive elements: `:focus-visible` mirrors hover states
- [ ] Enable "prefers-reduced-motion: reduce" in DevTools: all animations absent, final states visible
- [ ] Chrome (animation-timeline support): scroll-driven card borders and h2 underlines animate on entry
- [ ] Firefox/Safari (no animation-timeline): accent markers visible statically (no animation)

### DIFF
--- a/website/DESIGN.md
+++ b/website/DESIGN.md
@@ -135,3 +135,39 @@ Row 7b (accent on `--fleans-surface`) is dark-only; in light theme `--fleans-sur
 - Sidebar group labels (`.group-label .large`) and site title (`.site-title`) use Space Grotesk at weight 700 — verified against `@astrojs/starlight` 0.38.2 source.
 - `@fontsource` v5 ships `font-display: swap` in all generated CSS — no additional configuration needed.
 - IBM Plex Sans weight 500 is NOT imported — no Starlight 0.38.2 element uses it. All UI emphasis uses weight 600.
+
+## Motion
+
+CSS-only motion and hover effects — no JavaScript. Implemented in [#260](https://github.com/nightBaker/fleans/issues/260).
+
+### Principles
+
+1. **Motion is progressive enhancement.** Every animation is inside `@media (prefers-reduced-motion: no-preference)`. Users who prefer reduced motion see the final state instantly — no information is lost.
+2. **Hover effects require a pointer.** Hover rules live in `@media (hover: hover)`. `:focus-visible` equivalents are always active (outside the media query) so keyboard users get the same visual feedback.
+3. **No JavaScript, no intersection observers.** Scroll-driven animations use the CSS `animation-timeline: view()` spec behind an `@supports` guard — browsers without support see the static fallback (accent markers always visible).
+
+### Three moments
+
+| Moment | Technique | Timing |
+|---|---|---|
+| **Hero entrance** | `fade-up-clip` stagger (title → tagline → actions) + `clip-reveal-x` on logo | 320–420ms per element, 60–380ms stagger delay, `cubic-bezier(.2,.8,0,1)` (sharp-out) |
+| **Scroll accents** | Left-border `border-draw` on `.card`, underline `underline-draw` on `h2::after` | Tied to `animation-timeline: view()`, entry 0–40% |
+| **Hover / focus** | Sidebar lime `>` caret + 4px text shift; `.site-title` underline wipe; `.card` lift + hard lime shadow; primary action underline wipe; copy-button blinking lime dot | 120–180ms transitions |
+
+### Keyframes
+
+| Name | From | To | Notes |
+|---|---|---|---|
+| `fade-up-clip` | `translateY(8px)` + `clip-path: inset(100% 0 0 0)` | `translateY(0)` + `inset(0)` | Hero stagger |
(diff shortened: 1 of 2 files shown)


nightBaker (commented):
## Code Review -- CSS-only motion and hover effects (#260)

Clean implementation that faithfully translates the v2 design plan into working CSS. The a11y gates are correctly structured, keyframes match the plan, and the two-file scope is appropriate. A few observations below, none blocking.

### Approved -- no blockers found

### Observations (non-blocking)

1. **Hero action selectors: `:first-child`/`:nth-child(2)` vs `.action.primary`/`.action.secondary`.** The plan called for `.hero .action.primary` and `.hero .action.secondary` (verified selectors from Step 0). The implementation uses `.hero .action:first-child` and `.hero .action:nth-child(2)` instead. These are positionally equivalent today, but the class-based selectors are more robust against reordering or Starlight adding wrapper elements. Not blocking since the current DOM matches, but worth noting for future maintenance.

2. **Sidebar selector: `.sidebar-content a` vs `.sidebar-content a[href]`.** Plan specified `a[href]` to avoid matching any anchor-like elements without href. The implementation drops the `[href]` qualifier. In practice Starlight sidebar links always have href, so this is benign, but the attribute selector is slightly more defensive.

3. **`caret-slide` translateX value: `-8px` vs plan's `-4px`.** The keyframe uses `translateX(-8px)` while the design plan's keyframe inventory says `-4px`. The visual effect is that the caret slides in from further left. The difference is small and purely aesthetic -- the 8px version may actually feel snappier given the 140ms timing.

4. **Scroll-driven animation range: `entry 0% entry 40%` vs plan's `cover 10% cover 40%`.** The `entry` range is actually a better choice here -- it triggers as the element enters the viewport rather than during full coverage. This is a reasonable improvement over the plan.

5. **Copy-button `::after` vs plan's `::before`.** The plan specified `::before` for the lime dot, implementation uses `::after`. Functionally equivalent since the dot is absolutely positioned anyway. Just noting for plan-vs-implementation traceability.

6. **No `position: relative` on `.sidebar-content a` for focus-visible.** The hover block inside `@media (hover: hover)` sets `position: relative` on `.sidebar-content a`, which the absolutely-positioned `::before` caret needs. The `:focus-visible` block outside the hover query adds the `::before` pseudo-element with `position: absolute` but does not set `position: relative` on the anchor itself. On devices without hover capability (where the hover media query doesn't match), the caret's absolute positioning could be relative to a different containing block. This is the one item worth a quick manual check on a touch device with keyboard -- if the sidebar anchors already have `position: relative` from Starlight's styles, it's fine.

7. **Card transition + scroll-driven animation interaction.** The `.card` element has both a `transition` (for hover effects) and an `animation` (for scroll-driven border-draw, inside `@supports`). The `border-left` set in the scroll-driven block and the `border-color` transition in the hover block target different properties (border-left vs border-color), so they shouldn't conflict. Just noting the dual-animation surface for awareness.

8. **DESIGN.md Motion section** is well-written -- documents principles, keyframes, and accessibility approach. Matches the implementation accurately.

### Summary

The implementation correctly gates all motion on `@media (prefers-reduced-motion: no-preference)`, places hover rules inside `@media (hover: hover)`, keeps `:focus-visible` outside the hover media query, wraps scroll-driven animations in `@supports (animation-timeline: view())`, and defines all six planned keyframes. The DESIGN.md update is comprehensive. No JavaScript, no new dependencies. This is a solid, plan-faithful CSS-only PR.

nightBaker:
✅ All regression tests passed on HEAD 3746945 (904 passed / 1 skipped / 0 failed). Website build passes. Issue #260 moved to Review by Human.

PR #339, opened 2026-04-22
fix: merge compensation handler variables back to parent scope (#326)

### LINKED ISSUE #326: Bug: Compensation handler variable changes not merged to parent scope
## Bug Report

**Found during regression testing of #231 (PR #305)**

### Affected component
Compensation event execution in `WorkflowInstance` / `WorkflowExecution` aggregate.

### Expected behavior (per test plan and BPMN spec)
After a compensation walk completes, the main workflow scope should contain the merged results from all compensation handlers. For the `compensation-broadcast-process`:
```json
{ "hotelStatus": "cancelled", "flightStatus": "cancelled" }
```

### Actual behavior
Three separate variable scopes are left after compensation:
```
Scope A (main): {hotelStatus: "reserved", flightStatus: "cancelled"}
Scope B (pre-compensation): {hotelStatus: "reserved", flightStatus: "booked"}
Scope C (cancel_hotel isolated): {hotelStatus: "cancelled", flightStatus: "booked"}
```

**Root cause hypothesis:**
- `cancel_flight` (first handler in LIFO order) runs in the main scope and its changes propagate ✅
- `cancel_hotel` (second handler) receives a **cloned** scope (from the pre-compensation state) rather than the scope that `cancel_flight` updated, and its changes are NOT merged back to the main scope ❌

This means the second compensation handler's variable changes are silently dropped.

### Steps to reproduce
1. Deploy `tests/manual/24-compensation-event/compensation-broadcast.bpmn` (after fixing script syntax — see #325)
2. Start instance of `compensation-broadcast-process`
3. Wait for completion
4. Inspect `VariableStates` in `get_instance_state` — main scope will still show `hotelStatus: "reserved"`

### Impact
Compensation handlers that modify workflow variables produce inconsistent state. The last handler's changes may be lost, breaking any downstream logic that depends on compensation side-effects. Introduced in PR #305.

nightBaker:
## Summary

- After a compensation handler's child scope is seeded from the compensable activity's snapshot, overlay the parent scope's current variables so the handler also sees changes made by prior handlers in the same walk
- This fixes the bug where the second compensation handler would see stale variables and the root scope would lose the first handler's side-effects after the walk completed

## Root cause

In `AdvanceCompensationWalk()`, each handler's scope was seeded only from the immutable `snapshot.VariablesSnapshot`. Even though `AdvanceCompensationWalkIfHandlerCompleted()` correctly merged handler output back to the parent scope (line 655), the next handler's scope never picked up those parent-scope updates.

## Fix

Added a second `VariablesMerged` emit after the snapshot seed in `AdvanceCompensationWalk()` that overlays the parent scope's current variables onto the new handler scope. This ensures each handler starts with the snapshot as a base, then enriched with any parent-scope updates from prior handlers.

## Test plan

- [x] All 930 existing tests pass (253 application + 405 domain + 152 infrastructure + 116 persistence + 4 MCP)
- [x] Existing regression test `CompensationHandler_OutputVariables_ShouldPropagateToParentScope` passes — verifies handler_a sees handler_b's `flightStatus="cancelled"` and root scope shows both statuses as "cancelled"
- [ ] Manual test: `tests/manual/24-compensation-event/test-plan.md`

Fixes #326

### DIFF
--- a/src/Fleans/Fleans.Domain/Aggregates/WorkflowExecution.cs
+++ b/src/Fleans/Fleans.Domain/Aggregates/WorkflowExecution.cs
@@ -590,6 +590,13 @@ public class WorkflowExecution
         Emit(new ChildVariableScopeCreated(handlerVariablesId, parentVariablesId));
         Emit(new VariablesMerged(handlerVariablesId, snapshot.VariablesSnapshot));
 
+        // Overlay the parent scope's current variables so this handler sees changes
+        // made by previous compensation handlers in the same walk. Without this,
+        // each handler would only see the original snapshot and miss side-effects
+        // from earlier handlers that were already merged back to the parent scope.
+        var parentVariables = _state.GetVariableState(parentVariablesId).Variables;
+        Emit(new VariablesMerged(handlerVariablesId, parentVariables));
+
         // Spawn the handler activity
         var handlerInstanceId = Guid.NewGuid();
         Emit(new ActivitySpawned(


nightBaker (commented):
## Code Review — fix: merge parent scope variables into compensation handlers

Minimal, correct fix. The two-merge pattern (snapshot base + parent overlay) ensures each compensation handler in a LIFO walk sees variable changes from prior handlers.

**Verified:**
- `ChildVariableScopeCreated` creates the scope with parent linkage
- First `VariablesMerged` seeds from immutable snapshot (base)
- Second `VariablesMerged` overlays parent's current state (includes prior handler changes)
- Comment clearly explains the two-merge rationale
- Events are applied synchronously, so `_state.GetVariableState(parentVariablesId)` reflects the merge from the previous handler's completion

nightBaker (commented):
## Code Review — fix: merge compensation handler variables back to parent scope (#326)

Surgical 7-line fix in `AdvanceCompensationWalk()`. After seeding the handler's child scope from the compensable activity's snapshot, the fix overlays the parent scope's current variables so subsequent handlers see prior handlers' side-effects.

### Verified

- **Correctness:** The overlay order is correct — snapshot first (compensable activity's completion-time state), then parent scope on top (includes changes from previous handlers merged back via `AdvanceCompensationWalkIfHandlerCompleted`). This matches the CLAUDE.md design constraint: "After a handler completes successfully, its variable changes MUST be merged back into the enclosing scope before the next handler spawns."
- **Event sourcing:** Uses `Emit(new VariablesMerged(...))` consistent with the existing pattern. No direct state mutation.
- **Scope:** Only `WorkflowExecution.cs` changed. No infrastructure, API, or test changes needed — the existing `CompensationHandler_OutputVariables_ShouldPropagateToParentScope` test already validates this scenario.
- **CI:** All 930 tests pass.

PR #363, opened 2026-04-24
feat(editor): multi-tab BPMN editor (#352)

### LINKED ISSUE #352: tabs on editor page
allow user to open several bpmns in different tabs

nightBaker:
## Summary

Adds a tab bar to the `/editor` page. Users can keep up to 10 diagrams open at once, switch between them without losing unsaved edits, and have the set persisted across browser refreshes.

Closes #352.

## What changed

- **`Editor.razor`** — introduces a private `TabSession` model (`id`, `label`, `processKey`, `version?`, `bpmnXml`, `lastSavedXml`, `dirty`) tracked as `tabs: List<TabSession>` + `activeTabId`. Existing entry points (URL param, **New Diagram**, **Import BPMN**, **Download XML**, **Deploy**, process-ID edit) now operate on the active tab. Switching a tab calls `bpmn-js getXml` on the outgoing tab and `loadXml` on the incoming one — there is still only one `bpmn-js` modeler instance at a time.
- **`bpmnEditor.js`** — new interop helpers: `registerDirtyCallback`, `registerBeforeUnloadWarning`, `setDirtyTabsFlag`, `readStorage`, `writeStorage`. Dirty detection subscribes to bpmn-js `commandStack.changed` and notifies `.NET`; the `beforeunload` warning fires whenever any tab is dirty.
- **`app.css`** — adds `.editor-tab-bar` / `.editor-tab` / `.editor-tab--active` / `.editor-tab__close` / `.editor-tab__dirty` rules alongside the existing editor block.
- **Persistence** — tabs serialize to `localStorage` under key `fleans.editor.tabs.v1` after every state change (open / close / switch / dirty flip / deploy / rename). Restored on `OnAfterRenderAsync` before the URL-parameter entry path, so URL-param navigation still wins.
- **10-tab cap** — all "add tab" entry points gate on `tabs.Count < 10` and surface a dismissable `FluentMessageBar` error bar on overflow. Closing the last tab always opens a fresh blank one so the editor is never empty.
- **Confirm-close** — closing a dirty tab shows a FluentUI dialog ("Discard unsaved changes?"). Deploying a tab clears its dirty flag, because the deployed XML becomes the new saved snapshot.

## Docs + tests

- New guide at `website/src/content/docs/guides/editor.md` covering the reason for tabs, how to use them (open/switch/close/cap/persist/deploy), best practices, and known limitations. Sidebar entry added in `astro.config.mjs`.
- New manual test plan at `tests/manual/29-editor-tabs/test-plan.md` with six numbered scenarios (two-tab open, dirty + confirm-close, New Diagram + rename, 10-tab cap, refresh persistence, deploy clears dirty).
- Regression list in `CLAUDE.md` gets a new entry `32. Editor Tabs` pointing at the test plan, plus a short conventions note under **Code Conventions** covering the single-modeler / storage-key / cap invariants.

## Deviation from the design & plan

The approved plan called for **CallActivity double-click auto-open** as one of the tab-source entry points. That path needs additional bpmn-js interop wiring (subscribe to `element.dblclick` on `bpmn:CallActivity` + deployed-version lookup) and surface area worth its own PR. All other sources (New Diagram, Import BPMN, Workflows-list navigation) are fully wired. The limitation is called out explicitly in the Editor guide under **Known limitations**.

## How to test

Build + unit tests (nothing regresses; `Fleans.Web` has no unit tests):
```
dotnet build src/Fleans/Fleans.Web/Fleans.Web.csproj  # 0 warnings, 0 errors
```

Website build:
```
cd website && npm install && npm run build
```

Manual: follow `tests/manual/29-editor-tabs/test-plan.md` against `dotnet run --project Fleans.Aspire`.


nightBaker (commented):
## Code Review — feat(editor): multi-tab BPMN editor

Overall architecture matches the approved plan (single modeler + XML swap, localStorage-versioned key, 10-tab cap, confirm-close on dirty, docs + manual test plan added, CallActivity-dblclick deferred and disclosed). Three separate paths lose or corrupt tab XML, though — must fix before promotion.

### 🔴 Blocking

1. **`commandStack.changed` fires on `importXML` — every tab switch marks the incoming tab dirty.** bpmn-js emits `commandStack.changed` not just on user edits but also when `CommandStack.clear()` runs, which `importXML` does internally before loading new XML. That means every time `ActivateTab` calls `bpmnEditor.loadXml`, `OnModelerDirty` fires and flips `ActiveTab.Dirty = true` on the freshly-loaded tab. Effect: every tab becomes "dirty" the moment you switch to it, even if the user never touched it. The confirm-close dialog then triggers on tabs that have no real edits.

   Fix: track a `_suppressDirty` flag in `bpmnEditor.js` that `loadXml` sets to `true` around the `importXML` call and clears right after, and have the `commandStack.changed` handler early-return while it is set. (`newDiagram` goes through `loadXml` so it's covered by the same gate.)

2. **`NewDiagramTab` drops edits on the outgoing tab.** Every other "add a tab" path (`OnImportFileSelected`, `OpenFromWorkflow`) calls `SaveActiveTabXml()` before appending the new tab, but `NewDiagramTab` jumps straight to `AddBlankTab`. Because `AddBlankTab` calls `bpmnEditor.newDiagram` → replaces the modeler's current state with the blank canvas, whatever the user had in the outgoing tab since it was loaded is gone. Reproducer: open tab A, drag an element, click **New Diagram**, click tab A — A's diagram is back to the pre-edit state.

   Fix: `await SaveActiveTabXml();` at the top of `NewDiagramTab` before `AddBlankTab`. The same fix belongs inside `AddBlankTab` itself if you want it safe to call from any path, since the method is the one that actually mutates the modeler — leaving the save-outgoing responsibility to callers means future callers will trip over this again.

3. **URL-param entry clobbers restored tabs' XML after a refresh.** In `OnAfterRenderAsync`:

    ```
    dotNetRef = ...; bpmnEditor.init; registerDirtyCallback; registerBeforeUnloadWarning;
    editorInitialized = true;
    await RestoreTabsFromStorage();       // populates `tabs` and sets activeTabId from storage
    if (!string.IsNullOrEmpty(ProcessDefinitionKey))
        await OpenFromWorkflow(...);       // inside: await SaveActiveTabXml();
    ```

   At that `SaveActiveTabXml` call, the bpmn-js modeler has been init-ed but **no** restored tab has been loaded into it yet — the modeler is showing bpmn-js's default-empty diagram. `getXml` returns that default empty XML, and we then assign it over the previously-active restored tab's `BpmnXml`, overwriting the real content. Reproducer: open two workflows in tabs, refresh, land on `/editor/{keyA}/{v}` — tab B looks fine but tab A's contents are now the blank default.

   Fix: don't call `SaveActiveTabXml` in `OpenFromWorkflow` until the modeler is known to have loaded the current `ActiveTab` (i.e., it's safe after the first `ActivateTab`). Either (a) have the restore path `ActivateTab(activeTabId, saveOutgoing: false)` first so the modeler mirrors the active tab, then proceed to `OpenFromWorkflow`; or (b) teach `SaveActiveTabXml` to no-op until the modeler has been shown a tab at least once (e.g., a `_modelerHasContent` guard toggled on the first successful `loadXml`). Option (a) is cleaner since the rest of the code treats "active tab and modeler agree" as an invariant.

### 🟡 Important

4. **Persistence only fires on dirty-flip, not on ongoing edits.** `OnModelerDirty` skips when the tab is already dirty, so `PersistTabs` runs once per tab-dirty-transition but not on subsequent keystrokes. The design doc allows this ("serialize on every mutation" where edits flip-to-dirty is "a mutation"), but the net effect is: if I edit a tab heavily without switching, refresh the browser, I get back only the state at the first keystroke. A debounced persist on subsequent `commandStack.changed` events (e.g., every 2 seconds) would match the `beforeunload` contract better — right now the warning prompts the user to not close, but if they refresh after ignoring it, they still lose most edits.

5. **`TabSession.LastSavedXml` is dead code.** It's initialized, copied through DTO round-trips, and kept in sync on deploy, but no code ever *reads* it. The original design called for it as the basis of `dirty = bpmn !== lastSaved`, but the impl uses event-driven dirty tracking instead. Either wire it up (compute dirty on persist/switch as a safety net, or present a "revert to last saved" affordance) or remove it — storing unused XML blobs in localStorage is wasted quota.

### 🟢 Minor

6. `PerformCloseTab` calls `await PersistTabs()` and then `await UpdateDirtyFlagAndPersist()`. The latter also calls `PersistTabs()` — one redundant localStorage serialize per close. Collapse to a single call.

7. `.editor-tab-bar:empty { display: none; }` is a dead rule — the invariant in `PerformCloseTab` guarantees the list is never empty. Remove.

8. `AddBlankTab`'s `Untitled-N` counter uses `tabs.Count(t => t.Label.StartsWith("Untitled-"))` — if the user closes `Untitled-2` and creates another, the next gets `Untitled-2` again (reused). Not wrong, just slightly surprising in a session where tabs are named deterministically; a monotonic `_untitledSeq` would be clearer. Non-blocking.

Fixing items 1–3 will require a re-review since they change the mutation paths materially.

nightBaker:
Pushed d716865 addressing the three 🔴 Blocking and three 🟢 Minor items from the review:

1. ✅ `commandStack.changed` now suppressed during `importXML` via a `_suppressDirty` bracket in `loadXml` — tab switches no longer flip incoming tabs dirty.
2. ✅ `NewDiagramTab` now calls `SaveActiveTabXml()` before `AddBlankTab`, so outgoing edits are preserved when creating a new tab.
3. ✅ `OnAfterRenderAsync` now loads the restored active tab into the modeler before `OpenFromWorkflow` runs, so URL-param entry after a refresh no longer clobbers the previously-active restored tab's XML.
4. ✅ Removed dead `LastSavedXml` field and its DTO round-trip.
5. ✅ Collapsed redundant `PersistTabs` call in `PerformCloseTab`.
6. ✅ Dropped dead `.editor-tab-bar:empty` CSS rule.

Left as-is (worth a follow-up, not blocking this PR):

- 🟡 *Persistence only on dirty-flip, not ongoing edits.* Agreed, but fixing it well means adding a debounce path in `OnModelerDirty` that re-enables notifications after each persist — enough surface for its own change. Filed mentally for a follow-up.
- 🟢 `Untitled-N` counter reuses closed slots. Cosmetic; leaving until a monotonic `_untitledSeq` is clearly motivated.

CI will re-run on the new HEAD and the `bot-block` marker is invalidated by the new commit.

nightBaker (commented):
-

nightBaker:
Resolved merge conflicts with main (d427aa9). Only `CLAUDE.md` conflicted — main added a `32. Conditional Events` regression entry in parallel; renumbered Editor Tabs to #33, kept both. All other main changes auto-merged. `dotnet build Fleans.Web` stays green. CI will re-run on the merge commit.

nightBaker (commented):
## Code Review — feat(editor): multi-tab BPMN editor (post-merge)

Re-reviewed head `d427aa94` — a clean merge of `origin/main` into the feature branch. Only conflict was a regression-entry collision in `CLAUDE.md` (main added `32. Conditional Events`, branch added `32. Editor Tabs`); resolution keeps both, renumbered Editor Tabs to `#33`. All other files auto-merged.

Code changes on the feature branch were already reviewed and approved at commit `d716865f`:

- `_suppressDirty` bracket in `loadXml` + check in `registerDirtyCallback` — tab switches no longer flag incoming tabs dirty.
- `NewDiagramTab` now calls `SaveActiveTabXml()` before `AddBlankTab` — outgoing edits preserved.
- `OnAfterRenderAsync` loads the restored active tab before `OpenFromWorkflow` — URL-param entry after restore no longer clobbers.
- Dead `LastSavedXml` / redundant `PersistTabs` / `:empty` CSS rule removed.

Merge commit itself carries no behavioral changes. `dotnet build Fleans.Web` green post-merge.

🟡 Deferred (acknowledged): persistence only on dirty-flip, not ongoing edits — follow-up.
🟢 Deferred: `Untitled-N` counter reuses closed slots — cosmetic.

PR #369, opened 2026-04-24
fix(website): hero SVG strict-XML parse (#366)

### LINKED ISSUE #366: error getting svg on main landing page
https://nightbaker.github.io/fleans/hero-workflow-dark.svg

This page contains the following errors:
error on line 4 at column 9001: Opening and ending tag mismatch: g line 4 and svg
Below is a rendering of the page up to the first error.

nightBaker:
## Summary

`website/public/hero-workflow-{light,dark}.svg` carried 42 open `<g>` vs 21 close `</g>` tags on `main`, so strict XML consumers (direct browser view, feed readers, social-media previewers, sitemap crawlers) rejected them with `Opening and ending tag mismatch: g line 4 and svg`. The landing page rendered fine only because HTML's forgiving parser patched up the malformed tree at load time.

Closes #366.

## Root cause

`website/scripts/render-bpmn.mjs` post-processed the bpmn-js SVG output with string-regex overlay strippers:

```js
svg = svg.replace(/<[^>]+class="[^"]*djs-hit[^"]*"[^>]*>[\s\S]*?<\/[^>]+>/g, '');
// ...djs-outline, djs-dragger variants
```

bpmn-js 17 emits `djs-hit` / `djs-outline` / `djs-dragger` as self-closing `<rect .../>`. The regex's opening `<[^>]+>` absorbed the `/` of those self-closing tags (because `[^>]+` includes `/` as a non-`>` character), then the non-greedy `[\s\S]*?</[^>]+>` trailer consumed forward until the first available closing tag — typically the `</g>` that was supposed to close the parent `<g class="djs-element">`. Net: 21 `</g>` closers lost per file, one per stripped overlay cluster.

## What changed

- **`website/scripts/render-bpmn.mjs`** — replaced regex-based structural cleanup with DOM manipulation: call `viewer.saveSVG()` → parse with `DOMParser` → `querySelectorAll('.djs-hit, .djs-outline, .djs-dragger').forEach(el => el.remove())` + strip bpmn-font PUA text markers → re-serialize via `XMLSerializer` with the XML prolog + `<!-- created with bpmn-js -->` + SVG 1.1 `<!DOCTYPE svg …>` re-prepended. `postProcess` now does only theme recoloring + (pre-existing, no-op on root svg) viewBox padding.
- **`website/public/hero-workflow-{light,dark}.svg`** — regenerated. Output structure matches `origin/main` byte-for-byte beyond the removed overlay nodes (same root viewBox `155 35 759 330`, same 26 theme-stroke occurrences, same 9.2 KB file size).
- **`tests/manual/website/hero-bpmn-svg/test-plan.md`** (new) — 7 numbered checks: strict `xml.etree.ElementTree.parse`, `grep -c` for zero overlay classes, `head -3` literal-match for prolog+DTD, paired `<g>` balance (self-close-aware), direct-in-browser open (Firefox + Chrome), landing-page theme-switch render, regeneration reproducibility.
- **`CLAUDE.md`** — new guardrail bullet under `### Hero BPMN Diagram` explaining the `<[^>]+>`-absorbs-`/` failure mode so a future maintainer does not re-introduce regex-based structural edits; new numbered entry #2 under `## Website regression tests`.

## Verification

```
$ python3 -c "
import xml.etree.ElementTree as ET
ET.parse('website/public/hero-workflow-dark.svg')
ET.parse('website/public/hero-workflow-light.svg')
print('both files parse as strict XML')
"
both files parse as strict XML

$ grep -c "djs-hit\|djs-outline\|djs-dragger" website/public/hero-workflow-{light,dark}.svg
website/public/hero-workflow-light.svg:0
website/public/hero-workflow-dark.svg:0

$ head -3 website/public/hero-workflow-dark.svg
<?xml version="1.0" encoding="utf-8"?>
<!-- created with bpmn-js / http://bpmn.io -->
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">

$ cd website && npm run build
# 11 pages built, 0 errors
```

Side-by-side confirms the bug:

```
$ python3 -c "import xml.etree.ElementTree as ET
try: ET.parse('<origin/main copy>/hero-workflow-dark.svg')
except ET.ParseError as e: print(f'FAIL: {e}')"
FAIL: mismatched tag: line 4, column 8996
```

## Deviations from the approved Rev 2 design

Rev 2 called for Variant A — DOM cleanup in place followed by `viewer.saveSVG()` unchanged — on the premise that `saveSVG()` would reflect DOM mutations on the live canvas SVG. In practice on bpmn-js 17.11.1, `saveSVG()` reconstructs its output from the canvas model rather than serializing the live DOM, so live mutations are lost. Implementation switched to Variant B (from the analysis-review comment): round-trip the `saveSVG()` string through `DOMParser` → DOM manipulation → `XMLSerializer`, with the XML prolog and DOCTYPE re-prepended manually. Net output shape is identical to Variant A's target; pipeline is just one step longer.

Additionally observed during implementation: `document.querySelector('#canvas svg')` returns the `bjs-powered-by` logo badge, not the main drawing SVG (bpmn-js renders two SVGs under `#canvas`). Irrelevant now that the pipeline parses `saveSVG()`'s output instead of the live DOM, but noted in CLAUDE.md for anyone tempted to clone the live SVG directly in the future.

## How to test

1. `cd website && npm install && npx playwright install chromium` (one-time).
2. `cd website && npm run render-bpmn` — SVGs regenerate idempotently.
3. Run the 7 checks in `tests/manual/website/hero-bpmn-svg/test-plan.md`:
   - Strict XML parse (`python3 xml.etree.ElementTree.parse`) — both OK
   - `grep -c "djs-hit\|djs-outline\|djs-dragger"` — 0 both files
   - `head -3` — prolog + DOCTYPE present
   - Paired `<g>` balance — opens == closes
   - Open `/fleans/hero-workflow-{light,dark}.svg` directly in Firefox + Chrome — no parse-error banner
   - Landing page in both themes — hero renders cleanly
   - `npm run render-bpmn` twice — second run shows no `git diff`


nightBaker (commented):
## Code Review — fix(website): hero SVG strict-XML parse (#366)

Diff scope: 5 files, +160/-24. `website/scripts/render-bpmn.mjs` swaps regex-based structural cleanup for a DOMParser round-trip on `viewer.saveSVG()`'s output, then re-prepends the XML prolog + bpmn-js comment + SVG 1.1 DOCTYPE. Regenerated `hero-workflow-{light,dark}.svg` now parse as strict XML (verified in the PR body with `xml.etree.ElementTree.parse`); old files fail with `mismatched tag: line 4, column 8996`, exactly the #366 symptom. Manual test plan added at `tests/manual/website/hero-bpmn-svg/test-plan.md` (7 scriptable checks). CLAUDE.md carries the `<[^>]+>`-absorbs-`/` guardrail and #366 reference under `### Hero BPMN Diagram`, plus a numbered regression entry.

### Cross-checked

- ✅ Git workflow: feature branch (`fix/hero-svg-xml`), PR targets `main`, not a direct commit.
- ✅ Manual-test-plan rule (CLAUDE.md → "Manual Test Plans"): `tests/manual/website/hero-bpmn-svg/test-plan.md` covers the failure mode directly (strict XML) and adjacent invariants (prolog, overlay removal, paired `<g>` balance).
- ✅ Website regression list appended — entry #2 under `## Website regression tests`.
- ✅ Theme recoloring untouched; stripped overlay rects had `fill="none" stroke-opacity="0"`, so removal is color-neutral by construction.
- ✅ Prolog bytes match `origin/main`'s current files byte-for-byte (verified against the existing `hero-workflow-dark.svg` on main — same three-line header).
- ✅ Deviation from the approved Rev 2 design (Variant A → Variant B, forced by bpmn-js 17's `saveSVG()` rebuilding from canvas model rather than serializing the live DOM) is called out explicitly in the PR description.

### 🟡 Important (defensive, non-blocking for this PR)

1. **`DOMParser.parseFromString(rawSvg, 'image/svg+xml')` never throws on malformed input.** On parse failure it returns a document whose root is a `<parsererror>` element; `svgEl.querySelectorAll(…)` then silently no-ops, and the script writes an invalid output file to disk without exit-coding non-zero. Current bpmn-js 17 emits well-formed SVG, so this doesn't trigger today — but the point of #366 is that a future pipeline change can quietly produce malformed output, and the current guardrail is "the output is whatever `saveSVG()` + DOM manipulation produces, trust it." Recommend a one-line defense after the parseFromString call:
   ```js
   if (doc.querySelector('parsererror')) {
     throw new Error('DOMParser rejected saveSVG output: ' + doc.documentElement.textContent);
   }
   ```
   Happy to take this as a follow-up in the same PR if you'd like, or as a separate one-line fix — marking as non-blocking since `saveSVG()` is empirically well-formed in bpmn-js 17.11.1.

### 🟢 Minor

2. **The `viewBox` padding in `postProcess` is a pre-existing no-op on the root `<svg>`.** The output files show the first viewBox (a `<marker>`'s `0 0 20 20` inside `<defs>`) gets padded to `-20 -20 60 60`, not the root `<svg viewBox="155 35 759 330">`. Landing-page visual padding therefore comes entirely from CSS around the `` container. The test plan's "Known limitations" section acknowledges this honestly. Not in scope for #366 — but a future cleanup should either scope the regex to the root-svg viewBox (`svg.replace(/(<svg[^>]*?viewBox=")([^"]+)(")/, …)`) or delete the dead pad logic. Unrelated to this fix; noting for future reference.

3. **CLAUDE.md rule "visually inspect … in a browser"** could also cite the one-liner strict-XML check from `test-plan.md` step 1 as a faster pre-commit gate (`python3 -c "import xml.etree.ElementTree as ET; ET.parse('website/public/hero-workflow-dark.svg')"`). A visual check passes whether the file is strict-XML valid or not — HTML's forgiving parser absorbs the bug, which is what got us here originally. Scriptable check fails fast. Worth a one-line addition at the next touch of that bullet.

4. **`querySelectorAll('text')` walks the entire subtree.** Cheap on a 14-element BPMN fixture; no concern. Mentioning only because the loop body does `.trim()` + two `.charCodeAt(0)` checks for every text node — if the fixture grew to ~1000 text elements (implausible for a hero diagram), the filter could be tightened. Pure style; keep as-is.

### Verdict

No blocking issues. The 🟡 is defensive hardening for a scenario that doesn't currently occur. The three 🟢s are polish observations that don't affect this fix's correctness or CLAUDE.md compliance. Regenerated SVGs parse cleanly, old ones don't — #366's symptom is resolved.

OK to promote.

PR #355, opened 2026-04-23
feat(api): add opt-in JWT bearer authentication

nightBaker:
## Summary
- Add opt-in JWT bearer authentication to `Fleans.Api` via `Authentication:Authority` config key (#341)
- When `Authentication:Authority` is absent (default), API runs fully unauthenticated — zero behavior change
- When configured, all `/Workflow/*` endpoints require a valid `Authorization: Bearer <token>` header (401 otherwise)
- Health endpoints (`/health`, `/alive`) are exempt via `AllowAnonymous()` so load balancers/orchestrators can probe without credentials

## Changes
- `Fleans.Api.csproj` — add `Microsoft.AspNetCore.Authentication.JwtBearer` package
- `Program.cs` — conditional auth/authz service registration and middleware (authentication + authorization only wired when `Authentication:Authority` is set)
- `Extensions.cs` (ServiceDefaults) — `.AllowAnonymous()` on health check endpoint mappings
- `appsettings.json` — commented-out `Authentication` section as documentation
- `website/src/content/docs/reference/api.md` — authentication section with config table and Keycloak example
- `tests/manual/28-api-auth/test-plan.md` — manual test plan covering both auth-disabled and auth-enabled scenarios
- `CLAUDE.md` — regression test entry #31

## Test plan
- [x] `dotnet build` passes (0 errors)
- [x] `dotnet test` passes (930 tests, 0 failures)
- [ ] Manual: Scenario A (no auth config) — all endpoints work without tokens
- [ ] Manual: Scenario B (auth enabled) — unauthenticated requests get 401, valid tokens accepted, health endpoints remain open

Closes #341

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #341, opened 2026-04-22
oauth 2.0


nightBaker:
we need to support oauth 2.0 for authintification and authorization.
now we can use keycloack as auth provider, but it is should be configuraable with another providers in future.
management panel should also support oauth 2.0
also we need to support roles to have access to management pages and endpoints, may be we need to store all logged in users and use our management web to change roles of users by superadmin ? I want to know best practices it such kind of software which can be installed in premises

nightBaker:
## Design & Plan — OAuth 2.0 Authentication & Authorization

### Scope Analysis

This is a **large feature** spanning multiple layers. For on-premises software like Fleans, the standard approach is:

### Best Practices for On-Premises Auth

1. **Delegate authentication to an external IdP** — don't build your own login system. Support OIDC (OpenID Connect) which works with Keycloak, Azure AD, Okta, Auth0, Google, etc.
2. **Use ASP.NET Core's built-in auth middleware** — `AddAuthentication().AddOpenIdConnect()` + `AddAuthorization()` with policy-based authorization
3. **Roles via JWT claims** — the IdP issues tokens with role claims; Fleans just reads them. No need for Fleans to manage users/roles directly.
4. **Admin role management** — for on-prem, the superadmin manages roles in the IdP (Keycloak admin console). Fleans only needs to map IdP roles to authorization policies.

### Recommended Architecture

```
User → Fleans.Web (Blazor) → OIDC login → Keycloak/Azure AD/etc.
                                          ↓
                                    JWT with claims: {roles: ["admin", "operator"]}
                                          ↓
User → Fleans.Api → [Authorize(Policy="Admin")] → endpoint
```

### Proposed Sub-Issues

This should be split into incremental slices:

1. **Auth Slice 1: API Authentication** — Add OIDC/JWT bearer auth to `Fleans.Api`. Configurable via `appsettings.json` (Authority, ClientId, etc.). All existing endpoints get `[Authorize]`. Anonymous access preserved when auth section is absent (backward compatible).

2. **Auth Slice 2: Authorization Policies** — Define role-based policies (`Admin`, `Operator`, `Viewer`). Map to endpoint groups: admin endpoints require `Admin`, mutation endpoints require `Operator`, read endpoints allow `Viewer`.

3. **Auth Slice 3: Blazor Web Auth** — Add OIDC auth to `Fleans.Web` (Blazor Server). Login/logout flow, user display, role-gated UI components.

[...]

PR #380, opened 2026-04-26
feat(web): opt-in OIDC authentication for Management UI (#370)

### LINKED ISSUE #370: add authorization to fleans management
same way as #341

nightBaker:
## Summary

Implements **Slice 3 of #341** — opt-in OIDC Authorization Code flow with PKCE + cookie session for the `Fleans.Web` Blazor admin UI. Mirrors the opt-in pattern shipped for the API in #355: when no `Authentication` section is present, the UI runs unauthenticated exactly as today; setting `Authentication:Authority` AND `Authentication:ClientId` (single source of truth) flips on the full middleware stack.

Closes #370.

## What's in here

- `Security/AuthOptions.cs` — singleton record (`Enabled`, `Authority`, `ClientId`) consumed by `Routes.razor` to gate the cascading auth state and `<AuthorizeRouteView>`.
- `Program.cs` — D2 cookie + OIDC schemes, D3 fallback `[Authorize]`, D4 explicit `/dashboard` guard before `MapOrleansDashboard`, D6 login/logout endpoints with an inline `IsLocalUrl` open-redirect predicate (same shape as `IUrlHelper.IsLocalUrl`, expressed as a free static so the minimal-API delegate compiles without an MVC dependency), D10 keyed-Redis Data Protection key ring (multi-instance safe — configured via `KeyManagementOptions` because `PersistKeysToStackExchangeRedis` lacks an `IServiceProvider`-aware overload for keyed Redis), D12 forwarded-headers gated by `KnownProxies`/`KnownNetworks` allowlists.
- `Routes.razor` + `Security/RedirectToLogin.razor` — auth-aware router that wraps every page in `<AuthorizeRouteView>` when enabled (and falls back to today's plain `<RouteView>` when disabled).
- `NavMenu.razor` — `<AuthorizeView>` user chip with `Signed in as <preferred_username>` and an antiforgery-protected `Sign out` form.
- `Fleans.Aspire/Program.cs` — three optional `auth-authority`/`auth-client-id`/`auth-client-secret` parameters declared unconditionally with empty defaults, forwarded to `Fleans.Web` as env vars (D2a).
- `appsettings.example.jsonc` — documented config block (not loaded at runtime).
- `tests/manual/30-web-auth/` — full manual test plan + Keycloak Docker quickstart + importable realm JSON.
- `CLAUDE.md` — regression entry **#34** (last entry on `main` was `33. Editor Tabs`; `30-cancel-event/` is not on `main`, so the v7 Step-0 contingency does not fire).
- `website/.../authentication.md` — extended with a Management UI section covering config, Aspire wiring, behaviour when enabled, multi-instance Data Protection, reverse-proxy guidance, and pointers to the local-dev quickstart.

## Key decisions / deviations from v7 spec

- **Data Protection wiring** uses `KeyManagementOptions` + `RedisXmlRepository` rather than the `Func<IServiceProvider, IConnectionMultiplexer>` overload shown in v3 — that overload does not exist in `Microsoft.AspNetCore.DataProtection.StackExchangeRedis` 10.0 (only `IConnectionMultiplexer`/`Func<IDatabase>` overloads are public). The substituted form preserves the keyed-`orleans-redis` resolution semantics and is the canonical .NET 10 idiom for this case.
- **`KnownIPNetworks`** used instead of the obsolete `KnownNetworks` (per `ASPDEPR005`). Same effect, future-proof.

## How to test

- `dotnet build` — clean, 0 errors.
- `dotnet test` — 951/951 passing (1 pre-existing skip in `Fleans.Application.Tests` for `TimerEventSubProcess_NestedInsideSubProcess_CompletesScope`).
- Manual regression: follow [`tests/manual/30-web-auth/test-plan.md`](tests/manual/30-web-auth/test-plan.md). Quickstart for the local Keycloak in [`tests/manual/30-web-auth/keycloak-dev.md`](tests/manual/30-web-auth/keycloak-dev.md).

## Test plan

- [ ] `dotnet build` from `src/Fleans/` (clean, 0 errors)
- [ ] `dotnet test` from `src/Fleans/` (existing 950+ tests still green)
- [ ] Manual regression scenario 1 (auth disabled by default — no redirects, dashboard open, NavMenu has no chip)
- [ ] Manual regression scenario 2 (auth enabled — login round-trip preserves `?filter=active` deep link, NavMenu shows `Signed in as alice`)
- [ ] Manual regression scenario 3 (Orleans dashboard returns 302 → IdP when anonymous)
- [ ] Manual regression scenario 4 (every open-redirect attack input collapses to `/`; legitimate deep-link is honoured)
- [ ] Manual regression scenario 5 (bare `POST /Account/Logout` → 400; form-bound `Sign out` clears both schemes and re-challenges on next page load)
- [ ] Manual regression scenario 6 (`/health` and `/alive` stay anonymous when auth is on)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -172,6 +172,7 @@ The full regression suite is the union of every plan under `tests/manual/`. Each
 31. **API JWT Authentication** — `tests/manual/28-api-auth/test-plan.md`. Opt-in JWT bearer authentication; verifies API works unauthenticated by default, returns 401 when auth is configured and no token is provided, and accepts valid tokens.
 32. **Conditional Events** — `tests/manual/24-conditional-event/test-plan.md` (`conditional-event-test.bpmn`). Conditional intermediate catch event blocks until condition is true; conditional start event creates instances via evaluate-conditions API; conditional boundary event (interrupting) cancels host.
 33. **Editor Tabs** — `tests/manual/29-editor-tabs/test-plan.md`. Multi-tab BPMN editor in the Admin UI: open/switch/close tabs, dirty tracking with confirm-close dialog, 10-tab cap, `localStorage` persistence across refresh, `beforeunload` warning when any tab is dirty.
+34. **Management UI Authentication** — `tests/manual/30-web-auth/test-plan.md`. Opt-in OIDC for the Blazor Server admin UI; verifies (a) anonymous browse is allowed when no `Authentication` config is present, (b) `/dashboard` and any cascading-`AuthorizeView` page return 302 → IdP when auth is enabled, (c) login round-trip lands on the requested page (including `?query` parameters), (d) `/Account/Logout` is antiforgery-protected (bare POST is rejected, form-bound POST signs out and clears both schemes).
 
 > When adding a new manual test folder under `tests/manual/`, append a numbered entry here so the regression skill picks it up.
 
--- a/src/Fleans/Fleans.Aspire/Program.cs
+++ b/src/Fleans/Fleans.Aspire/Program.cs
@@ -10,6 +10,14 @@ var usePostgres = persistenceProvider.Equals("Postgres", StringComparison.Ordina
 // provider doesn't negotiate TLS. Disable to avoid health check failures (dotnet/aspire#13612).
 var redis = builder.AddRedis("orleans-redis").WithoutHttpsCertificate();
 
+// Authentication parameters (D2a) — declared unconditionally with empty defaults so the
+// AppHost contract is stable. When the operator does not supply values at run-time, the
+// downstream Web project sees empty strings for Authority/ClientId and falls into
+// auth-disabled mode (single source of truth lives in Fleans.Web/Program.cs).
+var authAuthority = builder.AddParameter("auth-authority", () => "");
+var authClientId = builder.AddParameter("auth-client-id", () => "");
+var authClientSecret = builder.AddParameter("auth-client-secret", () => "", secret: true);
+
 // Centralized Orleans configuration
 var orleans = builder.AddOrleans("cluster")
     .WithClustering(redis)
@@ -64,6 +72,9 @@ WithPersistence(
     builder.AddProject<Projects.Fleans_Web>("fleans-management")
         .WithReference(orleans.AsClient())
         .WaitFor(fleansSilo)
+        .WithEnvironment("Authentication__Authority", authAuthority)
+        .WithEnvironment("Authentication__ClientId", authClientId)
+        .WithEnvironment("Authentication__ClientSecret", authClientSecret)
         .WithReplicas(1),
     usePostgres, pg, sqliteConnectionString);
 
--- a/src/Fleans/Fleans.Web/Components/Layout/NavMenu.razor
+++ b/src/Fleans/Fleans.Web/Components/Layout/NavMenu.razor
@@ -1,3 +1,5 @@
+@using Microsoft.AspNetCore.Components.Forms
+
 <div class="navmenu">
     <FluentAppBar>
         <FluentAppBarItem Href="/workflows"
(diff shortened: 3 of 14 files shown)


nightBaker (commented):
## Code Review — feat(web): opt-in OIDC for Management UI (#370 / Slice 3 of #341)

The change is well-shaped: opt-in gating is a single `authEnabled` boolean derived from `Authority`+`ClientId`, the regression entry (`CLAUDE.md` #34), `tests/manual/30-web-auth/`, and `website/.../authentication.md` are all in this PR (so the project's "docs is part of done" rule from CLAUDE.md is satisfied), and the spicy bits (open-redirect, dashboard guard, multi-instance Data Protection, forwarded-headers allowlist) are individually reasoned. No blocking issues — flagging two 🟡 items to verify during the manual regression and a few 🟢 nits.

### 🟡 Important — verify during scenario 5 (manual regression)

- `app.MapPost("/Account/Logout", (HttpContext ctx) => Results.SignOut(...))` ([Program.cs:212-217](src/Fleans/Fleans.Web/Program.cs)) doesn't bind any form data, so it doesn't carry `IAntiforgeryMetadata`. In .NET 8/9 the `UseAntiforgery` middleware short-circuits with 400 on endpoints that bind form data (`IFormCollection`, `[FromForm]`, Razor Components SSR). A bare minimal-API POST that just takes `HttpContext` is **not automatically validated** even though the NavMenu form sends `__RequestVerificationToken`. Scenario 5 (`tests/manual/30-web-auth/test-plan.md`) explicitly expects 400 for the bare `curl -X POST /Account/Logout --cookie ...`. If that test returns 200 + sign-out, the cleanest fix is to bind the form so the middleware kicks in:

  ```csharp
  app.MapPost("/Account/Logout", ([FromForm(Name="__RequestVerificationToken")] string _, HttpContext ctx) =>
      Results.SignOut(...));
  ```

  or call `IAntiforgery.ValidateRequestAsync(ctx)` explicitly inside the handler. Note: `o.Cookie.SameSite = SameSiteMode.Lax` already mitigates real cross-site CSRF (the auth cookie won't be sent on a cross-site POST), so this isn't a fresh-attack vector — but the documented behavior should match reality, otherwise the regression entry in CLAUDE.md becomes a known-bad assertion.

### 🟡 Important — forwarded-headers gated on `authEnabled`

- `app.UseForwardedHeaders(...)` ([Program.cs:128-145](src/Fleans/Fleans.Web/Program.cs)) only registers when auth is on. An operator who runs `Fleans.Web` behind a TLS-terminating reverse proxy with auth disabled (still a supported config per the opt-in design) will see `Request.Scheme == "http"`, and `app.UseHttpsRedirection()` will issue a 301 → `https://internal-host/...`, which the proxy re-routes back, looping. Forwarded-headers is a deployment concern, not an auth concern — consider lifting the block out of the `if (authEnabled)` gate (still controlled by the `KnownProxies` / `KnownNetworks` allowlists, so an empty list = no behavior change). If you want to keep this scoped, add a one-line note in `website/.../authentication.md` and `appsettings.example.jsonc` that "auth disabled assumes Fleans.Web is *not* behind a proxy that terminates TLS".

### 🟢 Minor

- `o.Scope.Add("openid"); o.Scope.Add("profile")` ([Program.cs:90-91](src/Fleans/Fleans.Web/Program.cs)) — both are already in the OIDC handler's default scope collection, so these two lines are redundant. Harmless (most IdPs dedupe), but trim if you want a cleaner diff. `email` is the only scope that genuinely needs to be added.
- `MapPost("/Account/Logout", ...)` lacks `.AllowAnonymous()`. With the `RequireAuthenticatedUser` fallback policy, an anonymous POST is bounced through the IdP just to land on a sign-out endpoint — odd UX (mostly a curiosity since the realistic caller is always authenticated). Adding `.AllowAnonymous()` makes the intent explicit.
- The two `if (authEnabled)` blocks in `Program.cs` (services + middleware) total ~95 lines and would read more cleanly extracted into `ServiceCollectionExtensions.AddFleansWebAuth` / `WebApplicationExtensions.UseFleansWebAuth`. Style preference, follow-up sized.
- `tests/manual/30-web-auth/test-plan.md` Scenario 2 step 2 mentions port `7124`, but `Fleans.Web`'s actual launch port (per `Properties/launchSettings.json` / Aspire forward) is worth double-checking — every other regression plan in `tests/manual/` uses `https://localhost:7140` (the API). If the Web port is different on this machine, the universal prerequisite paragraph at the top of CLAUDE.md's "Regression tests" section reads "Web UI reachable at `https://localhost:7140`", which is now a documentation drift risk.
- D34 entry in `CLAUDE.md` says `/Account/Logout` is antiforgery-protected — see 🟡 #1 above; promote/demote depending on the regression run.

### Compliance with CLAUDE.md

- ✅ Manual test plan present and registered (`#34` in the regression list)
- ✅ Website docs updated in same PR (`website/src/content/docs/reference/authentication.md`)
- ✅ Activity/grain conventions N/A here (no domain code touched)
- ✅ Aspire is the startup project; the auth parameters are declared unconditionally with empty defaults so the AppHost contract stays stable

OK to promote pending the manual regression — particularly Scenario 5 (which validates the antiforgery claim) and Scenario 6 (verifies `/health` and `/alive` — already wired with `.AllowAnonymous()` in `Fleans.ServiceDefaults/Extensions.cs:108,114`).

PR #382, opened 2026-04-26
feat(streaming): pluggable Kafka stream provider for Orleans Streams (#372)

### LINKED ISSUE #372: add pluggable queue for orleans streams
pluggable as authentication
as example we can use kafka https://github.com/OrleansContrib/Orleans.Streams.Kafka

nightBaker:
## Summary

Closes #372. Adds an opt-in Kafka backend behind the existing `AddFleanStreaming()` switch so production silos can publish workflow domain events through a durable broker while local dev keeps the in-memory provider. Default is unchanged — `dotnet run --project Fleans.Aspire` boots in-memory streams with no Kafka container.

- New `Fleans.Streaming.Kafka` project — `IQueueAdapterFactory` / `IQueueAdapter` / `IQueueAdapterReceiver` over `Confluent.Kafka` 2.6.1, with `[GenerateSerializer]` batch container and idempotent client-side `AdminClient.CreateTopicsAsync` ensure step.
- `FleanStreamingExtensions.AddFleanStreaming` grows a `"kafka"` case (case-insensitive); error message lists both providers.
- Aspire `WithStreaming` helper that conditionally provisions a `fleans-kafka` resource and forwards `Fleans__Streaming__Provider` / `Fleans__Streaming__Kafka__Brokers` onto the silo. Reads `FLEANS_STREAMING_PROVIDER` (default `Memory`).
- Env-gated integration tests (`FLEANS_KAFKA_TESTS=1`) over Testcontainers-Kafka — smoke + at-least-once-after-restart + topic-ensure.
- Manual test plan #35 (`tests/manual/35-kafka-streaming/`) with a chained-script-task BPMN fixture and a silo-kill scenario.
- Docs page `website/src/content/docs/reference/streaming.md` covering the provider switch, delivery contract, topic provisioning, and the v1-is-plaintext-only scope statement.

## Delivery semantics

**At-least-once** with consumer-side idempotency. Offsets are committed AFTER handler success, so a silo crash mid-handle replays the event on resume. `WorkflowInstance.CompleteActivity` / `FailActivity`'s stale-callback guard (`if (!State.HasActiveEntry(activityInstanceId)) return;`) makes the replay a logged no-op. The streaming docs page calls this out explicitly so the guard isn't removed by accident.

## Scope statement

v1 ships plaintext brokers only — no SASL, no TLS, no mTLS. Production-managed Kafka (Confluent Cloud, MSK, Aiven, Redpanda Cloud) **requires** the production-auth follow-up that exposes `SecurityProtocol` / `SaslMechanism` / `SaslUsername` / `SaslPassword` / mTLS keys. Topic partition / replication tunables and the trivial `WorfklowEvaluateConditionEventHandler` typo cleanup are also follow-up issues filed alongside this PR.

## Test plan

- [ ] CI build passes across the solution (Memory mode unchanged).
- [ ] Memory-mode test suites (`Fleans.Application.Tests`, `Fleans.Domain.Tests`) green — Kafka tests should report `Inconclusive` without `FLEANS_KAFKA_TESTS=1`.
- [ ] On a machine with Docker: `FLEANS_KAFKA_TESTS=1 dotnet test src/Fleans/Fleans.Application.Tests` runs `KafkaStreamProviderIntegrationTests` + `KafkaTopicEnsureTests` green against Testcontainers-Kafka.
- [ ] Manual test #35 passes: `FLEANS_STREAMING_PROVIDER=Kafka dotnet run --project Fleans.Aspire`, deploy `kafka-streams.bpmn`, start instance, kill silo between tasks, restart → both script tasks complete with `first` = `"done"` and `second` = `"done"`.
- [ ] Aspire dashboard's `fleans-core` env tab shows `Fleans__Streaming__Provider=Kafka` and a non-empty `Fleans__Streaming__Kafka__Brokers` only when the env var is set.
- [ ] Default `dotnet run --project Fleans.Aspire` (no env var) provisions no Kafka container.


nightBaker (commented):
-

nightBaker (commented):
## Code Review — feat(streaming): pluggable Kafka stream provider for Orleans Streams (#372)

Scope, surface, tests, and docs all match the v3 design + Analysis Review v3 sign-off. Default path is genuinely zero-infra (no Kafka container when `FLEANS_STREAMING_PROVIDER` is unset), the `kafka` switch is purely additive in `FleanStreamingExtensions`, and the new manual test #35 / docs page / regression-list entry all line up with the CLAUDE.md "documentation + test plan are part of done" rules. BPMN fixture has its `<bpmndi:BPMNDiagram>` block (CLAUDE.md fixture rule), and the new `reference/streaming.md` is picked up by `autogenerate: { directory: 'reference' }` in `astro.config.mjs` so no sidebar wiring needed.

The at-least-once contract + consumer-side idempotency story is internally consistent: offsets are committed in `MessagesDeliveredAsync` (after the handler returns), and the docs page calls out the `WorkflowInstance.HasActiveEntry` stale-callback guard as the load-bearing invariant — good defensive documentation.

### 🟡 Important

1. **Dead-code expression in `KafkaQueueAdapter.QueueMessageBatchAsync` key derivation** — `src/Fleans/Fleans.Streaming.Kafka/KafkaQueueAdapter.cs:71-73`:
   ```csharp
   var key = streamId.ToString().AsSpan().ToArray() is { } _
       ? System.Text.Encoding.UTF8.GetBytes(streamId.ToString())
       : Array.Empty<byte>();
   ```
   This pattern-matches a non-null `char[]` (always succeeds), throws away the `char[]`, then calls `streamId.ToString()` a second time inside `GetBytes`. The whole expression is equivalent to `var key = System.Text.Encoding.UTF8.GetBytes(streamId.ToString());`. Two wasted allocations per produce on the hot path, and a confusing read for the next maintainer. Worth fixing before this lands — it's a one-line edit and removes a code smell that will get cargo-culted.

### 🟢 Minor (no need to block on these)

1. `_lastDeliveredOffset` in `KafkaQueueAdapterReceiver.cs:18` is written on every consume (line 84) but never read. Either drop the field or wire it into the commit log so the warning on commit failure has useful context.
2. `KafkaTopicNaming.TopicForQueue(options, queueId)` ignores `options` entirely — the XML comment explains why ("use queueId.ToString() verbatim"), but a `_ = options;` discard or removing the parameter would make the intent obvious at the call site.
3. `ProducerConfig` sets `EnableIdempotence = false` with `Acks.All`. That's consistent with the at-least-once contract you've documented, but enabling Confluent's idempotent producer is essentially free and would shrink the *produce-side* duplicate window without changing the handler-side guarantee. Worth a follow-up issue (next to the SASL/TLS one) if you want to tighten the duplicate rate later.
4. `KafkaBatchContainer.SequenceToken` falls back to `new KafkaSequenceToken(0, 0)` when `Token` is null. In practice `Token` is always set by the receiver before handlers see the batch, so this is defensive — fine to leave, but a comment explaining when the fallback can fire would prevent future readers from staring at it.

PR #367, opened 2026-04-24
feat(infra): Fleans.Worker project + Fleans:Role config (#356)

nightBaker:
## Summary

Foundation for the Core/Worker role split (#356): carve the script and condition StatelessWorker grain implementations into a new `Fleans.Worker` class-library project, introduce a `Fleans:Role` configuration key that stamps the role into the silo name, and keep the grain interfaces in `Fleans.Application` so callers don't need a Worker reference.

Closes #356.

## Scope (and what's deferred)

This PR implements the **structural refactor** from the approved v4 plan — everything needed for the `Combined` mode to keep working byte-for-byte identically to today, and the scaffolding `Fleans:Role=Worker` / `Fleans:Role=Core` need to even start up. The **runtime placement filtering** layer from the plan — `WorkerPlacementStrategy` / `CorePlacementStrategy` / their directors + the `[CorePlacement]` attributes on a dozen core grains — is **deferred to a follow-up PR**.

Rationale for the split: the placement-strategy layer requires Orleans-10-specific `IPlacementDirector` / `IManagementGrain.GetDetailedHosts` verification that benefits from being its own surface-area-limited change. The structural move sitting at HEAD first makes the follow-up's diff focused on placement logic and its tests, not a hundred lines of project restructuring mixed in. CLAUDE.md records this explicitly: today's split is structural, not placement-filtered.

## What changed

- **New project `src/Fleans/Fleans.Worker/`** — class library, `net10.0`, referenced by `Fleans.Api`, `Fleans.Application.Tests`, `Fleans.Infrastructure.Tests`, and added to `Fleans.sln`.
- **Moved two grain implementations** (via `git mv` so history is preserved):
  - `Fleans.Application/Scripts/ScriptExecutorGrain.cs` → `Fleans.Worker/Scripts/ScriptExecutorGrain.cs` (namespace `Fleans.Worker.Scripts`)
  - `Fleans.Application/Conditions/ConditionExpressionEvaluatorGrain.cs` → `Fleans.Worker/Conditions/ConditionExpressionEvaluatorGrain.cs` (namespace `Fleans.Worker.Conditions`)
- **`Fleans.Api/Program.cs`** — reads `Fleans:Role` (values `Core`, `Worker`, `Combined`, case-insensitive, default `Combined`; unknown values throw) and stamps `SiloOptions.SiloName = "{role}-{machine}-{guid}"`.
[...]

PR #384, opened 2026-04-27
feat(#357): pluggable custom-task framework — v12 (sub-issue A)

nightBaker:
## Summary

Closes part of #357 — sub-issue A (framework scaffold + catalog grain), per design v12.

This PR is a **hard-reset** of `feature/custom-task-framework` onto current `main`, replacing the v1–v9 provider-grain registry approach with the v12 architecture: per-plugin handler grains that subscribe to a shared event stream and self-filter by task type, plus a Core-side catalog grain that tracks which plugins are alive on which Worker silos.

Sub-issues B (REST plugin), C (UI page), and D (author guide) are tracked as follow-ups in #357.

### Architecture (v12)

```
[BPMN] <serviceTask type="..."> + <zeebe:ioMapping>
   ↓ BpmnConverter
[CustomTaskActivity]
   ↓ ExecuteAsync — publishes ExecuteCustomTaskEvent
[ExecuteCustomTaskEvent on stream "events/ExecuteCustomTaskEvent"]
   ↓ ↓ ↓ (fan-out, one activation per concrete plugin handler class)
[RestCallerHandler]   [EmailSenderHandler]   …  ← each [WorkerPlacement]
  taskType=rest-call    taskType=email
  if (!match) return;   if (!match) return;
   ↓ resolve inputs (MappingResolver), run plugin work, project outputs
   ↓ workflowInstance.CompleteActivity / FailActivity
```

Catalog flow:
```
[Worker silo boot]
  CustomTaskPluginRegistrar (ILifecycleParticipant<ISiloLifecycle>, stage=Active)
    foreach descriptor in DI:
      catalog.Register(taskType, siloName, schema)        ← one-shot, no heartbeat

[Core silo]
  CustomTaskCatalogGrain ([CorePlacement])
    RegisterTimer(period=30s, ReconcileWithMembership)    ← polls IManagementGrain
                                                            and drops entries for
                                                            departed silos
    GetAll() / Get(taskType) → /custom-tasks API
```

### Files

**Domain (`Fleans.Domain`)**
- `Activities/CustomTaskActivity.cs`, `Activities/Mappings.cs`
- `Errors/CustomTaskFailedActivityException.cs` — typed failure with `(code, message)`
- `Events/ExecuteCustomTaskEvent.cs`
- `GrainStorageNames.cs` — added `CustomTaskCatalog` constant

**Application (`Fleans.Application`)**
- `CustomTasks/ICustomTaskCatalogGrain.cs` + `CustomTaskCatalogGrain.cs` — `[CorePlacement]` singleton with 30s reconcile timer polling `IManagementGrain.GetDetailedHosts()`; keeps entries for silos in `{Joining, Active, ShuttingDown}`, drops the rest
- `CustomTasks/CustomTaskRegistration.cs` + `CustomTaskCatalogEntry.cs`
- `CustomTasks/MappingResolver.cs` — input/output mapping path walker
- `Events/WorkflowEventsPublisher.cs` — switch case for `ExecuteCustomTaskEvent`
- `Logging/WorkflowLoggingContext.cs` — visibility raised to `public` for cross-assembly use from the Worker handler base

**Worker (`Fleans.Worker`)**
- `CustomTasks/CustomTaskHandlerBase.cs` — abstract base; subclasses inherit `[WorkerPlacement]` + `[ImplicitStreamSubscription("events")]`. Subclasses override `TaskType` and `ExecuteAsync(...)` only; the base owns Orleans stream subscription, taskType filter, error path, and `IWorkflowInstanceGrain.CompleteActivity` / `FailActivity` callbacks (line-for-line analogue of `WorkflowExecuteScriptEventHandler`)
- `CustomTasks/CustomTaskPluginRegistrar.cs` — `ILifecycleParticipant<ISiloLifecycle>` at `ServiceLifecycleStage.Active` with bounded retry (2s, 5s, 15s) for catalog-cold-start resilience; warn-and-continue on terminal failure
- `CustomTasks/CustomTaskPluginDescriptor.cs`
- `CustomTasks/CustomTaskServiceCollectionExtensions.cs` — `services.AddCustomTaskPlugin<THandler>(taskType, displayName?, parameterSchemaJson?)` DI extension

**Infrastructure (`Fleans.Infrastructure`)**
- `Bpmn/BpmnConverter.cs` — `<serviceTask type="...">` and `<zeebe:taskDefinition type="...">` routing to `CustomTaskActivity`; `<zeebe:ioMapping><zeebe:input/output>` parsing with target identifier validation; mapping-source grammar validation at deploy time; `__response` reserved on output targets

**Api (`Fleans.Api`)**
- `Controllers/CustomTasksController.cs` — `GET /custom-tasks` (list), `GET /custom-tasks/{taskType}` (single)
- `Fleans.ServiceDefaults/DTOs/CustomTaskCatalogEntryDto.cs`

**Tests (39 new)**
- `Fleans.Domain.Tests/CustomTaskActivityTests.cs` — 4 (ctor / property / null / exception state-mapping)
- `Fleans.Application.Tests/CustomTasks/MappingResolverTests.cs` — 8 (carry-over: all four source forms + error paths)
- `Fleans.Application.Tests/CustomTasks/CustomTaskCatalogGrainTests.cs` — 6 (register / upsert / aggregate / case-insensitive / unknown-type)
- `Fleans.Infrastructure.Tests/BpmnConverter/CustomTaskRoutingTests.cs` — 21 (carry-over: routing + io-mapping deploy-time validation)

**Docs**
- `website/src/content/docs/concepts/custom-tasks.md` — rewritten for v12
- `website/astro.config.mjs` — sidebar entry under Concepts
- `README.md` — Service Task row in the BPMN elements table
- `CLAUDE.md` — "How to Add a Custom Task Plugin" section + regression list entry #37

**Manual test plan**
- `tests/manual/37-custom-task-framework/stub-custom-task.bpmn` + `test-plan.md` — three scenarios (unregistered plugin → manual complete-activity; registered plugin → end-to-end auto-complete; multi-silo reconcile drops stopped silo)

## Key decisions

1. **No edits to `WorkflowExecution.cs` / `WorkflowInstance.Infrastructure.cs`.** `CustomTaskActivity` mirrors `ScriptTask`'s publish-event-and-return-empty pattern verbatim. No new `IExecutionCommand` record; engine aggregate plays no role in the dispatch path.
2. **Per-plugin grain class, not a shared `ICustomTaskExecutorGrain`.** Each plugin has its own concrete grain. Orleans single-activation on a non-`[StatelessWorker]` `[ImplicitStreamSubscription]` grain gives one activation per plugin class cluster-wide; the in-base `if (TaskType != ...) return;` filter handles fan-out cheaply.
3. **Catalog uses cluster-membership polling, not heartbeats.** `IManagementGrain.GetDetailedHosts(onlyActive: false)` polled every 30 s — same precedent as `WorkerPlacementDirector.cs:36` and `CorePlacementDirector.cs:36`. No async-enumerable subscriptions, no fire-and-forget Tasks on the grain, no `IsAlive()` callbacks (which couldn't reach the implicit-subscription activation by `Guid` anyway).
4. **Registrar at `ServiceLifecycleStage.Active`, not `IHostedService`.** Guarantees the silo has finished joining the cluster before the Register call lands.
5. **Reconcile keeps `Joining|Active|ShuttingDown`.** Tolerant of in-flight silo state transitions; freshly-registered entries from a still-`Joining` peer survive the reconcile pass.

## Deviations from v12 design

1. **Catalog state is in-memory only in v1**, not `IPersistentState`. After a Core silo restart, the catalog repopulates as Worker silos restart and re-register. EF-backed persistence (so the UI is correct immediately after Core restart) is a v2 follow-up — the per-entry table + EF migrations for both SQLite and PostgreSQL providers are sub-issue A2 work; the in-memory model is sufficient for the dev/Aspire path where Core and Worker share a process. **Documented in the docs page and in the catalog grain's class summary.**
2. **End-to-end TestCluster integration test deferred to manual test plan #37.** Orleans assembly-part discovery for test-assembly-defined plugin handlers is non-trivial (the runtime walks production assemblies referenced by the silo, not the test assembly that defines stub handlers). The framework PR's coverage is 39 unit tests + the BPMN routing tests + the catalog grain tests; the dispatch path is structurally identical to `WorkflowExecuteScriptEventHandler` which already has full coverage.

## How to test

- [x] `dotnet test` — full suite passes (~600 tests, 0 failures, 5 pre-existing skips).
- [x] `cd website && npm run build` — clean (17 pages including new `/concepts/custom-tasks/`).
- [ ] Manual test plan #37 — three scenarios listed in `tests/manual/37-custom-task-framework/test-plan.md`.

## Follow-ups (filed against #357)

- **Sub-issue B** — `Fleans.Plugins.RestCaller` end-to-end with a Kestrel mock-server hermetic test.
- **Sub-issue C** — `CustomTasks.razor` admin page consuming `GET /custom-tasks`.
- **Sub-issue D** — plugin author guide on the website.
- **A2** — EF-backed catalog persistence (SQLite + Postgres migrations) so the UI is immediately correct after Core silo restart.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,6 +75,22 @@ The splash page (`website/src/content/docs/index.mdx`) loads an interactive Thre
    - **Fail tests:** error state set with correct code/message (500 for generic Exception, 400 for BadRequestActivityException), failed activity still transitions to next activity, no variables merged on failure
 4. Update the BPMN elements table in `README.md`
 
+## How to Add a Custom Task Plugin
+
+A *custom task* is a `<bpmn:serviceTask type="…">` whose execution is supplied by a user-written plugin grain on a Worker silo. Use this — **not** "Add a New BPMN Activity" — when the new behavior is plugin-shaped (REST call, email, custom external system).
+
+1. Add a new project (e.g. `Fleans.Plugins.MyThing`) referencing `Fleans.Worker`. Inside it, write a class deriving from `Fleans.Worker.CustomTasks.CustomTaskHandlerBase`. Override `TaskType` and `ExecuteAsync(...)`. The base class carries `[ImplicitStreamSubscription("events")]` and `[WorkerPlacement]` — subclasses inherit both.
+2. Throw `Fleans.Domain.Errors.CustomTaskFailedActivityException(int code, string message)` from `ExecuteAsync` to fail with a typed error; any other thrown exception fails with code 500.
+3. The plugin's `ExecuteAsync` returns an `IDictionary<string, object?>`. Output mappings (`<zeebe:output source="=__response.body" target="…"/>`) walk that dictionary.
+4. Expose a DI extension method on the plugin assembly:
+   ```csharp
+   public static IServiceCollection AddMyThingPlugin(this IServiceCollection services) =>
+       services.AddCustomTaskPlugin<MyThingHandler>(taskType: "my-thing", displayName: "My Thing");
+   ```
+   Plugin authors who want their plugin to live in the catalog UI must call this from the Worker silo's host registration.
+5. Tests: write unit tests for the plugin's logic (call `ExecuteAsync` directly with stub inputs); end-to-end TestCluster integration is exercised by manual test plan #37 once a real plugin ships.
+6. Documentation: update `website/src/content/docs/concepts/custom-tasks.md` with the plugin's parameter schema and any limitations.
+
 ## How to Add a New API Endpoint
 
 Add it to `Fleans.Api/Controllers/WorkflowController.cs`. DTOs go in `Fleans.ServiceDefaults/`.
@@ -179,6 +195,7 @@ The full regression suite is the union of every plan under `tests/manual/`. Each
 34. **Management UI Authentication** — `tests/manual/30-web-auth/test-plan.md`. Opt-in OIDC for the Blazor Server admin UI; verifies (a) anonymous browse is allowed when no `Authentication` config is present, (b) `/dashboard` and any cascading-`AuthorizeView` page return 302 → IdP when auth is enabled, (c) login round-trip lands on the requested page (including `?query` parameters), (d) `/Account/Logout` is antiforgery-protected (bare POST is rejected, form-bound POST signs out and clears both schemes).
 35. **Kafka Streaming Provider** — `tests/manual/35-kafka-streaming/test-plan.md` (`kafka-streams.bpmn`). Opt-in Kafka stream provider via `FLEANS_STREAMING_PROVIDER=Kafka`; verifies (a) Aspire dashboard provisions a `fleans-kafka` resource and forwards `Fleans__Streaming__Provider`/`Fleans__Streaming__Kafka__Brokers` env vars to the silo, (b) chained-script-task workflow completes after the silo is killed and restarted between tasks (at-least-once delivery), (c) the client-side `AdminClient.CreateTopicsAsync` ensure step auto-creates topics with the configured prefix.
 36. **Cancel Event (Transaction Cancellation)** — `tests/manual/30-cancel-event/test-plan.md` (`cancel-transaction.bpmn`). Cancel End Event inside a Transaction triggers: active scope activities cancelled, Cancel Boundary Event fires, recovery flow runs to completion. Verifies transaction outcome is Cancelled.
+37. **Custom Task Framework** — `tests/manual/37-custom-task-framework/test-plan.md` (`stub-custom-task.bpmn`). `<serviceTask type="...">` parses to `CustomTaskActivity`; with no plugin registered the activity stays Active indefinitely (manual `complete-activity` API call resumes it); registered plugin via `services.AddCustomTaskPlugin<T>()` claims the event, runs, and completes the activity; `GET /custom-tasks` reflects registered/dropped plugins as silos join/leave.
 
 > When adding a new manual test folder under `tests/manual/`, append a numbered entry here so the regression skill picks it up.
 
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ For now, next elements are implemented
 | **Activities**       |                                                                             |             |
 | Task                 | A single unit of work.                                                      |     [x]     |
 | Script Task          | A task that executes an inline script.                                       |     [x]     |
+| Service Task         | A task whose execution is supplied by a custom plugin (`<serviceTask type="…">`); plain `<serviceTask>` (no `type`) is a no-op. | [x] custom-task framework |
 | Sub-Process          | A group of tasks that are treated as a single unit.                         |     [x]     |
 | Call Activity        | A type of sub-process that calls another process.                            |     [x]     |
 | User Task            | A task that requires human interaction with claim lifecycle.                 |     [x]     |
--- a/src/Fleans/Fleans.Api/Controllers/CustomTasksController.cs
+++ b/src/Fleans/Fleans.Api/Controllers/CustomTasksController.cs
@@ -1,0 +1,42 @@
+using Fleans.Application.CustomTasks;
+using Fleans.ServiceDefaults.DTOs;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Fleans.Api.Controllers;
+
+/// <summary>
+/// Surfaces the <see cref="ICustomTaskCatalogGrain"/> contents to the management UI.
+/// </summary>
+[ApiController]
+[Route("custom-tasks")]
+public sealed class CustomTasksController : ControllerBase
+{
+    private readonly IGrainFactory _grainFactory;
+
+    public CustomTasksController(IGrainFactory grainFactory)
+    {
+        _grainFactory = grainFactory;
+    }
+
+    /// <summary>Returns one entry per registered task type with the silos hosting it.</summary>
+    [HttpGet]
+    public async Task<IReadOnlyList<CustomTaskCatalogEntryDto>> GetAll()
+    {
+        var catalog = _grainFactory.GetGrain<ICustomTaskCatalogGrain>(0);
+        var entries = await catalog.GetAll();
+        return entries
+            .Select(e => new CustomTaskCatalogEntryDto(e.TaskType, e.DisplayName, e.ParameterSchemaJson, e.SiloNames))
+            .ToList();
+    }
+
+    /// <summary>Returns the entry for one task type, 404 if no plugin claims it.</summary>
+    [HttpGet("{taskType}")]
+    public async Task<ActionResult<CustomTaskCatalogEntryDto>> Get(string taskType)
+    {
+        var catalog = _grainFactory.GetGrain<ICustomTaskCatalogGrain>(0);
+        var entry = await catalog.Get(taskType);
+        if (entry is null)
+            return NotFound();
+        return new CustomTaskCatalogEntryDto(entry.TaskType, entry.DisplayName, entry.ParameterSchemaJson, entry.SiloNames);
+    }
+}
--- a/src/Fleans/Fleans.Application.Tests/CustomTasks/CustomTaskCatalogGrainTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/CustomTasks/CustomTaskCatalogGrainTests.cs
@@ -1,0 +1,94 @@
+using Fleans.Application.CustomTasks;
+
+namespace Fleans.Application.Tests.CustomTasks;
+
+[TestClass]
+public class CustomTaskCatalogGrainTests : WorkflowTestBase
+{
+    private ICustomTaskCatalogGrain GetCatalog() =>
+        Cluster.GrainFactory.GetGrain<ICustomTaskCatalogGrain>(0);
+
+    [TestMethod]
+    public async Task Register_NewEntry_AppearsInGetAll()
+    {
+        var catalog = GetCatalog();
+
+        await catalog.Register(new CustomTaskRegistration("rest-call", "REST Caller", null, "worker-A"));
+
+        var entries = await catalog.GetAll();
+        var rest = entries.SingleOrDefault(e => e.TaskType == "rest-call");
+        Assert.IsNotNull(rest);
+        Assert.AreEqual("REST Caller", rest!.DisplayName);
+        Assert.HasCount(1, rest.SiloNames);
+        Assert.AreEqual("worker-A", rest.SiloNames[0]);
+    }
+
+    [TestMethod]
+    public async Task Register_SameTaskTypeAndSilo_IsIdempotent()
+    {
+        var catalog = GetCatalog();
+
+        await catalog.Register(new CustomTaskRegistration("rest-call", null, null, "worker-A"));
+        await catalog.Register(new CustomTaskRegistration("rest-call", "Updated Display", null, "worker-A"));
+
+        var entries = await catalog.GetAll();
+        var rest = entries.Single(e => e.TaskType == "rest-call");
+        Assert.HasCount(1, rest.SiloNames);
+        Assert.AreEqual("Updated Display", rest.DisplayName);
+    }
+
+    [TestMethod]
+    public async Task Register_SameTaskTypeDifferentSilos_AggregatesIntoOneEntry()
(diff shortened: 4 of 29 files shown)


nightBaker (commented):
-

nightBaker (commented):
## Code Review — feat(#357): pluggable custom-task framework (sub-issue A)

Solid scaffold. Architecture matches the v3 design (per-plugin grain interfaces + registry + stream-routed `ExecuteCustomTaskEvent`). Records, `[GenerateSerializer]`, `List<T>` collections, `[LoggerMessage]` source generators, and `[CorePlacement]` placement all line up with CLAUDE.md.

### 🟡 Important
1. **At-least-once semantics for `OnNextAsync`.** The handler catches everything and either calls `FailActivity` (swallowing) or rethrows when `FailActivity` itself fails — by design for retry. But Orleans implicit-stream subscriptions can redeliver, so a successful provider call followed by a failing `CompleteActivity`/`FailActivity` will be retried with a fresh provider grain (`GetGrain(..., ActivityInstanceId)` is the same activation, but `ExecuteAsync` runs again). This is in line with CLAUDE.md's "each activity instance executes at most once" guarantee from the engine's perspective, but plugin authors may execute external side effects twice. Worth a sentence in `concepts/custom-tasks.md` (or the `ICustomTaskCallProvider` XML doc) telling plugin authors to make `ExecuteAsync` idempotent — not a code change.

2. **Plugin-side mutation of `variables` is silently dropped.** `WorkflowExecuteCustomTaskEventHandler.OnNextAsync` copies inputs into a fresh `resolved` dict, calls the provider with both `resolved` and `variables`, then only projects `resolved` (via `OutputMappings`) into the `outputs` ExpandoObject sent to `CompleteActivity`. If a plugin author writes directly into the passed-in `variables` ExpandoObject expecting persistence, those mutations vanish. The `ICustomTaskCallProvider` doc-comment hints at this (says to write `resolved["__response"]`) but doesn't explicitly forbid mutating `variables`. Worth a one-liner: *"`variables` is a read-only snapshot; persist outputs by writing to `resolved` and defining `<zeebe:output>` mappings."*

3. **`__response` is a magic string used in two places.** `BpmnConverter.ParseOutputMapping` rejects `target == "__response"` with a literal, and `concepts/custom-tasks.md` documents it as the convention for plugins. Promoting that to a `public const string` somewhere in `Fleans.Application/CustomTasks` (or `Fleans.Domain/Activities`) would prevent the strings from drifting apart in sub-issues B/C.

### 🟢 Minor
- `MappingResolver.WalkPath`: `IDictionary<string, object?>` and `ExpandoObject` branches are functionally identical (an ExpandoObject *is* an `IDictionary<string, object?>`). The first `if` already matches an ExpandoObject — the second `else if` is dead code in practice. Worth a quick local check; if true, drop the second branch and simplify.
- `CustomTaskFailedActivityException` stores `_code` and `_message` as `private readonly` Orleans-id fields. Other `ActivityException` subclasses (e.g. `BadRequestActivityException`) tend to use auto-properties — purely stylistic, no impact on behaviour.
- `BpmnConverter.ResolveServiceTaskType` accepts both a bare `type=` attribute and `<zeebe:taskDefinition type="…"/>`. The website docs only show the Zeebe form. Consider documenting (or dropping) the bare-attribute path so authors don't get into the habit of using it and then trip over the editor not round-tripping it.

### Files reviewed
20 changed files, 991 additions / 2 deletions. Spot-checked the diff against CLAUDE.md sections on records, `[GenerateSerializer]`, `List<T>`, `[LoggerMessage]`, Core/Worker split, and "Each activity instance executes at most once" — no violations.

src/Fleans/Fleans.Worker/Events/Handlers/WorkflowExecuteCustomTaskEventHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            if (!_registry.TryGetGrainInterface(item.TaskType, out var grainInterface))
+                throw new CustomTaskFailedActivityException(400, $"No provider registered for task type '{item.TaskType}'");
+
+            var resolved = new Dictionary<string, object?>();
+            foreach (var im in item.InputMappings)
+                resolved[im.Target] = MappingResolver.Resolve(im.Source, (IDictionary<string, object?>)variables);
+
+            var provider = (ICustomTaskCallProvider)_grainFactory.GetGrain(grainInterface!, item.ActivityInstanceId);
+            await provider.ExecuteAsync(resolved, variables);
```
nightBaker: custom task should not be executed in core
service task should produce event with customServiceTaskName and parameters 
customServiceTask (as worker) listen (on worker silo) and execute custom code, then complete or fail service task

nightBaker (commented):
-

nightBaker (commented):
## Code Review — feat(#357): pluggable custom-task framework (sub-issue A)

The maintainer left a substantive architectural concern at `WorkflowExecuteCustomTaskEventHandler.cs:72` after the prior `bot-approve` was emitted. That concern stands and overrides the prior approval — re-emitting as `bot-block` so the worker stops attempting promotion.

### 🔴 Blocking

1. **Custom task code runs on Core silos — violates the role split documented in CLAUDE.md.**
   - `WorkflowExecuteCustomTaskEventHandler` is annotated `[CorePlacement]` (file head), so its stream subscription and the subsequent `provider.ExecuteAsync(...)` call (lines ~70-73) execute on a Core silo.
   - CLAUDE.md "Core / Worker role split" section explicitly says: *"When adding a new worker-type grain (e.g. the upcoming REST-call service task), put the implementation in `Fleans.Worker` and keep the interface next to the caller in `Fleans.Application`."* This PR does the opposite — the interface is in `Fleans.Application`, but no implementation moves to `Fleans.Worker`, and the dispatching grain is pinned to Core.
   - Per the maintainer's own thread: *"custom task should not be executed in core; service task should produce event with customServiceTaskName and parameters; customServiceTask (as worker) listen (on worker silo) and execute custom code, then complete or fail service task."*
   - **Required fix:** invert the topology. The `WorkflowExecuteCustomTaskEventHandler` (Core-side) should only emit `ExecuteCustomTaskEvent` to the stream and not call providers. A separate worker-silo grain (e.g. `Fleans.Worker.CustomTaskExecutorGrain`, `[StatelessWorker]` or with `WorkerPlacement`) subscribes to the same stream, resolves the provider, runs `ExecuteAsync`, then calls back via `IWorkflowInstanceGrain.CompleteActivity` / `FailActivity`. This mirrors how `ScriptExecutorGrain` and `ConditionExpressionEvaluatorGrain` are split today (interfaces in `Fleans.Application`, implementations in `Fleans.Worker`).
   - Until this is fixed, `Combined`-role silos appear to work but a real `Core`/`Worker` split (the production topology documented in CLAUDE.md) executes user-supplied custom code in the coordinator silo, which is the exact failure mode the role split was introduced to prevent.

### 🟡 Important

1. **Unbounded scope on a single PR.** 991 additions across 20 files for "sub-issue A" of #357 is wide. After the topology fix above, consider splitting into: (a) event + handler dispatch (Core side), (b) executor grain in `Fleans.Worker`, (c) registry + mapping resolver, (d) BPMN converter integration + tests, (e) docs. Smaller PRs make the architectural review tractable; the current size is a contributing factor to the topology error slipping through the prior review.

2. **`MappingResolver` is invoked with implicit provider mutation.** Comment at lines 77-78 says *"the provider mutated `resolved` to add `__response`"* — the contract that providers may mutate the input dictionary in place is implicit. If the worker-silo executor is reached over the stream, that contract has to survive serialization (the `resolved` dict is local to the executor; whatever the provider writes back has to land on `outputs` via `OutputMappings`, not via shared-dict mutation). Worth either documenting the post-execute mutation contract on `ICustomTaskCallProvider` or replacing it with an explicit `ProviderResult` return type.

### 🟢 Minor

1. `BpmnConverter.cs` adds 156 lines for routing; if those are mostly dictionary lookups against the registry, a small `CustomTaskRoutingResolver` class would isolate them and keep `BpmnConverter` from accreting more responsibilities.
2. Docs — `website/src/content/docs/concepts/custom-tasks.md` is added (good), but should call out the Core/Worker placement of the executor explicitly so readers understand where `ExecuteAsync` runs. Easy to add once the topology fix lands.

Re-request review after the executor moves to `Fleans.Worker` and the Core handler stops calling `provider.ExecuteAsync` directly.

src/Fleans/Fleans.Worker/Events/Handlers/WorkflowExecuteCustomTaskEventHandler.cs
nightBaker: Addressed in `95315ab`.

Moved `WorkflowExecuteCustomTaskEventHandler` from `Fleans.Application/Events/Handlers/` to `Fleans.Worker/Events/Handlers/` and switched `[CorePlacement]` → `[WorkerPlacement]`. `WorkerPlacementDirector` (PR #367) routes activations to silos with `worker-` or `combined-` prefix, so on a real Core/Worker split the dispatcher only activates on Worker silos and `provider.ExecuteAsync(...)` runs there.

Mirrors the existing `ScriptExecutorGrain` split (interface in `Fleans.Application`, impl in `Fleans.Worker`).

Provider impls also need `[WorkerPlacement]` — without it Orleans falls back to `RandomPlacement` and can land a provider grain on a Core silo even when the dispatcher is on Worker. Documented this on the `ICustomTaskCallProvider` XML doc, in CLAUDE.md ("How to Add a Custom Task Provider"), and on the website concepts page (replaced the "no placement filter" v2 stub with a real Placement section).

Side-effect: `Fleans.Application` now exposes internals to `Fleans.Worker` via `InternalsVisibleTo` so the moved handler can keep using `WorkflowLoggingContext` (kept `internal` rather than promoting the whole logging surface to `public`).

Full test suite ran green: 984 passed, 0 failed, 1 pre-existing skip.

nightBaker:
## Re-review request — `95315ab`

Addressed the bot-block on the unresolved thread.

### What changed

| Change | Where |
|---|---|
| Moved `WorkflowExecuteCustomTaskEventHandler` from `Fleans.Application` → `Fleans.Worker` | `src/Fleans/Fleans.Worker/Events/Handlers/WorkflowExecuteCustomTaskEventHandler.cs` (rename, 92% similarity) |
| `[CorePlacement]` → `[WorkerPlacement]` on the dispatch grain | same file |
| Documented `[WorkerPlacement]` requirement for provider impls | `ICustomTaskCallProvider` XML doc, `CLAUDE.md` step 2, `website/.../custom-tasks.md` |
| `InternalsVisibleTo("Fleans.Worker")` so handler keeps using `WorkflowLoggingContext` | `Fleans.Application.csproj` |

### Why

`WorkerPlacementDirector` (PR #367) filters compatible silos to those whose `SiloName` carries `worker-` or `combined-` prefix. With the move + placement attribute swap, a Core-only silo will not host the dispatcher and will not invoke `provider.ExecuteAsync(...)`. On Combined-role silos the filter is a no-op (every silo is `combined-*`), so single-process dev/test remains unchanged.

The provider grains themselves still need `[WorkerPlacement]` — `IGrainWithGuidKey` defaults to `RandomPlacement`, which would let a provider activate on Core even with the dispatcher on Worker. Made this explicit in the docs.

### Bot review tradeoffs (from the prior block)

- **🔴 Blocking — fixed** as above.
- **🟡 Important — not in this commit:**
  - PR scope. Splitting an in-flight large PR retroactively isn't practical; happy to keep a smaller-PR rule in mind for the v2 follow-up.
  - `MappingResolver` provider mutation contract. The mutation now happens inside the Worker grain (post-move), so the cross-silo serialization concern doesn't apply — the dictionary is always local to the dispatcher's process. Documenting an explicit `ProviderResult` return type is a separate refactor; left for a v2 follow-up.
- **🟢 Minor — partly addressed:**
  - Docs Placement section rewritten to explicitly cover where `ExecuteAsync` runs.
  - `BpmnConverter` extraction filed as #429 (was already on the Backlog before this commit).

### Verification

- `dotnet build` — clean (warnings unchanged from baseline; all pre-existing).
- `dotnet test` — 984 passed, 0 failed, 1 pre-existing skip.

CI rollup will re-run on this push. Re-requesting review.

PR #438, opened 2026-05-01
feat(#434): EF-backed catalog persistence

### LINKED ISSUE #434: Sub-issue A2 of #357: EF-backed catalog persistence
## Sub-issue A2 of #357 — EF-backed catalog persistence

The custom-task catalog grain (`CustomTaskCatalogGrain`, landed in #384) uses in-memory state in v1. After a Core silo restart, the catalog is empty until Worker silos restart and re-register — meaning the management UI shows "no plugins" until each worker reboots, even when workers are healthy.

This sub-issue closes that gap by wiring `IPersistentState` for the catalog with an EF-backed `IGrainStorage`, mirroring the precedent set by `ConditionalStartEventRegistryGrain` / `EfCoreConditionalStartEventRegistryGrainStorage`.

### Scope

- Add a `CustomTaskCatalogState` record (or per-entry table) plus `EfCoreCustomTaskCatalogGrainStorage : IGrainStorage` in `Fleans.Persistence`.
- DbSet wiring in `FleanCommandDbContext` + model configuration in `FleanModelConfiguration`.
- New EF Core migration in `Fleans.Persistence.PostgreSql/Migrations/Command/`. SQLite is auto-handled via `EnsureCreated()`.
- DI registration: `services.AddKeyedSingleton<IGrainStorage>(GrainStorageNames.CustomTaskCatalog, ...)` in `Fleans.Persistence/DependencyInjection.cs`.
- `CustomTaskCatalogGrain` switches from the in-memory dictionary to `IPersistentState<CustomTaskCatalogState>`; on `OnActivateAsync`, the existing `ReconcileWithMembership()` runs immediately so persisted entries from departed silos are dropped on first activation (already implemented; just needs to fire before the first read).
- Tests: `Fleans.Persistence.Tests/EfCoreCustomTaskCatalogGrainStorageTests.cs` covering round-trip + remove paths.

### Acceptance

- [ ] After Core silo restart, `GET /custom-tasks` returns the last-known catalog within milliseconds — no need to wait for worker restarts.
- [ ] Reconcile still drops entries whose silo is no longer in cluster membership (timer fires every 30s; runs once on activation).
- [ ] Postgres migration applies cleanly via `MigrateAsync()` on `Fleans.Api` startup.
- [ ] Existing v1 in-memory tests (`CustomTaskCatalogGrainTests`) all still pass.

### References

- Framework reference: `website/src/content/docs/concepts/custom-tasks.md` — note explicitly mentions EF persistence as a v2 follow-up.
- Precedent: `Fleans.Persistence/EfCoreConditionalStartEventRegistryGrainStorage.cs`.
- Catalog grain: `Fleans.Application/CustomTasks/CustomTaskCatalogGrain.cs`.

nightBaker:
## Summary

Closes #434 (sub-issue A2 of #357) — replaces the in-memory dictionary in `CustomTaskCatalogGrain` with `IPersistentState<CustomTaskCatalogState>` backed by an EF Core grain-storage provider. After Core silo restart, the catalog reactivates with persisted entries within milliseconds; the existing 30 s reconcile timer drops anything whose silo left the cluster while Core was down.

### Domain

- **New** `Fleans.Domain/States/CustomTaskCatalogState.cs` — Orleans-serializable state record holding `List<CustomTaskCatalogRowState>`; `Upsert` + `RemoveWhere` helpers. The **"Row" suffix** differentiates per-row state from the public `CustomTaskCatalogEntry` DTO (aggregated by `TaskType` with silo list) — explicit per the v1 critique #3.

### Persistence

- **New** `EfCoreCustomTaskCatalogGrainStorage` — mirrors `EfCoreConditionalStartEventRegistryGrainStorage` line-for-line: read = load all rows; write = diff existing-vs-new keyed by `(TaskType, SiloName)`; clear = truncate.
- `DbSet<CustomTaskCatalogRowState> CustomTaskCatalogEntries` on `FleanCommandDbContext`.
- Entity configured in `FleanModelConfiguration` with composite PK on `(TaskType, SiloName)`; `ParameterSchemaJson` as default text type (`TEXT` on SQLite, `text` on PostgreSQL).
- DI registration in `Fleans.Persistence/DependencyInjection.cs` as keyed `IGrainStorage` at `GrainStorageNames.CustomTaskCatalog`.
- **Postgres migration** `AddCustomTaskCatalog` (timestamp `20260501162400`) + Designer + model-snapshot update. SQLite picks up the new table automatically via `EnsureCreated()`.

### Application

- `CustomTaskCatalogGrain` switches from in-memory `_entries` dictionary to `IPersistentState<CustomTaskCatalogState>`. `Register` / `ReconcileWithMembership` call `WriteStateAsync` only when state actually changed — no-op writes avoided.
- `OnActivateAsync` now **reconciles before registering the timer** so persisted-but-stale entries get dropped before the first request lands.
- `ParameterSchema` ↔ JSON conversion via `System.Text.Json` (default settings; no `TypeNameHandling` since these are first-party POCO records).
- **Skip-and-warn on malformed schema JSON** (EventId 9325, Warning level): return `null` schema instead of throwing — UI shows the plugin without parameter widgets rather than breaking the entire catalog. v1 critique #2 picked option (a).

### Test infrastructure

- `WorkflowTestBase.cs` adds `.AddMemoryGrainStorage(GrainStorageNames.CustomTaskCatalog)` alongside the other in-memory storages so existing `CustomTaskCatalogGrainTests` (8 tests) continue to pass against the now-persistent state machinery. v1 critique #1.

### Tests (7 new)

`Fleans.Persistence.Tests/EfCoreCustomTaskCatalogGrainStorageTests.cs`:
- Empty-database read returns no record.
- Round-trip preserves all fields including null DisplayName / ParameterSchemaJson.
- Upsert same key updates not duplicates.
- Remove drops the row.
- Multi-silo same-task-type round-trips both rows.
- Clear drops all rows.

All 7 new tests pass; existing 8 `CustomTaskCatalogGrainTests` in `Application.Tests` continue to pass against memory-backed `IPersistentState`.

### Docs + CLAUDE.md

- `concepts/custom-tasks.md` "Catalog & liveness" section dropped the "(v1)" qualifier; documents EF-backed persistence + composite PK + JSON serialization. Notes the boot-time write-volume implication for large fleets (v1 critique #5).
- `CLAUDE.md` persistence section gains a bullet describing the grain-storage name + table layout for future contributors.

## Key decisions

1. **JSON-column for ParameterSchema** — single text field, serialized via `System.Text.Json`. The catalog never queries inside the schema; per-parameter relational queries have no current consumer. Matches how plugin authors think about the schema (one record per plugin).
2. **System.Text.Json with default settings** — no `TypeNameHandling`-equivalent risk; these are first-party POCO records, not external untrusted data like response bodies. (Workflow variables use Newtonsoft+ExpandoObject because they're dynamic; that doesn't apply here.)
3. **Skip-and-warn on malformed JSON** (EventId 9325) — most resilient option. Single bad row returns `null` schema; rest of the catalog continues to work. UI shows the plugin without parameter widgets. Future schema migrations don't break existing rows.
4. **Reconcile-on-activate is now load-bearing** — same call site as before, but with persistence the catalog wakes up holding entries the cluster's no longer keeping. The reconcile pass drops them in milliseconds, before the first request lands.
5. **Postgres migration written by hand**, not via `dotnet ef migrations add`. Equivalent output (single CreateTable with composite PK), avoids the global-tool prereq + project-reference dance.

## How to test

- [x] `dotnet build` — clean.
- [x] `dotnet test --filter "FullyQualifiedName~CustomTask"` — 48 tests pass (4 Domain + 21 BPMN routing + 16 Application + 7 Persistence; existing tests + 7 new persistence tests).
- [x] `cd website && npm run build` — 18 pages clean.
- [ ] Manual: stop the Aspire stack with at least one plugin registered; confirm `CustomTaskCatalogEntries` table holds rows (SQLite: open `fleans.db`; Postgres: `SELECT * FROM "CustomTaskCatalogEntries"`); restart Aspire; confirm `GET /custom-tasks` returns the entry within ~1 s without waiting for Worker silo restart.

## Deviations from v1 design

None. Every recommendation from v1 + the review (option (a) skip-and-warn for JSON errors, "Row" suffix on the state record, WorkflowTestBase wiring, dotnet-ef prereq note, write-volume callout) is implemented as specified.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -228,6 +228,7 @@ Two providers: **SQLite** (default, local dev) and **PostgreSQL** (production/lo
 - **Migrations live per-provider:** `Fleans.Persistence.Sqlite/Migrations/Command/` and `Fleans.Persistence.PostgreSql/Migrations/Command/`. Only command-context migrations are maintained (command and query share the same database).
 - **Provider packages:** `Fleans.Persistence.Sqlite` and `Fleans.Persistence.PostgreSql` — each registers a `RelationalModelCustomizer` subclass via `ReplaceService<IModelCustomizer>` for provider-specific model tweaks (e.g., SQLite stores `DateTimeOffset` as string; PostgreSQL uses native `timestamptz`).
 - **Adding a new provider:** Create a new `Fleans.Persistence.<Provider>` project, implement a `<Provider>ModelCustomizer : RelationalModelCustomizer`, add an `Add<Provider>Persistence()` extension, generate initial migrations, and wire into host `Program.cs` files.
+- **Custom-task catalog persistence:** `CustomTaskCatalogGrain` uses `IPersistentState<CustomTaskCatalogState>` keyed by `GrainStorageNames.CustomTaskCatalog`, backed by `EfCoreCustomTaskCatalogGrainStorage` and the `CustomTaskCatalogEntries` table (composite PK on `(TaskType, SiloName)`; `ParameterSchemaJson` stored as a JSON text column via `System.Text.Json`). Test clusters must register memory grain storage for this name in `WorkflowTestBase` alongside the other registries.
 
 ## Things to Know
 
--- a/src/Fleans/Fleans.Application.Tests/WorkflowTestBase.cs
+++ b/src/Fleans/Fleans.Application.Tests/WorkflowTestBase.cs
@@ -186,6 +186,7 @@ public abstract class WorkflowTestBase
                 .AddMemoryGrainStorage(GrainStorageNames.UserTasks)
                 .AddMemoryGrainStorage(GrainStorageNames.ConditionalStartEventListeners)
                 .AddMemoryGrainStorage(GrainStorageNames.ConditionalStartEventRegistry)
+                .AddMemoryGrainStorage(GrainStorageNames.CustomTaskCatalog)
                 .AddCustomStorageBasedLogConsistencyProviderAsDefault()
                 .UseInMemoryReminderService()
                 .ConfigureServices(services =>
--- a/src/Fleans/Fleans.Application/CustomTasks/CustomTaskCatalogGrain.cs
+++ b/src/Fleans/Fleans.Application/CustomTasks/CustomTaskCatalogGrain.cs
@@ -1,24 +1,23 @@
+using System.Text.Json;
 using Fleans.Application.Placement;
+using Fleans.Domain;
+using Fleans.Domain.States;
 using Microsoft.Extensions.Logging;
 using Orleans.Runtime;
 
 namespace Fleans.Application.CustomTasks;
 
 /// <summary>
-/// In-memory catalog of custom-task plugin registrations announced by Worker silos.
-/// Membership-reconciliation timer prunes entries whose silo has left the cluster.
-///
-/// State is intentionally ephemeral in this v1: on Core silo restart the catalog is
-/// rebuilt as Worker silos restart and re-register. EF-backed persistence is a v2
-/// follow-up — design v12 originally specified <c>IPersistentState</c>, deferred here
-/// because the per-entry table + migrations dwarf the rest of sub-issue A and the
-/// in-memory model is sufficient for the dev/aspire path where Core and Worker silos
-/// share a process.
+/// Catalog of custom-task plugin registrations announced by Worker silos.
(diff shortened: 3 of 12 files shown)


nightBaker:

Issue #432, opened 2026-05-01
Sub-issue C of #357: Management UI typed parameter editor


nightBaker:
## Sub-issue C of #357 — Management UI: typed parameter editor for custom tasks

When a workflow author drops a `<serviceTask>` into the BPMN editor and selects a custom-task `type`, the properties panel today renders only generic key/value rows. This sub-issue surfaces the typed `CustomTaskParameterSchema` (landed in #430) into the editor so each parameter renders with the correct primitive widget.

### Scope

- Fetch `GET /custom-tasks` from the management UI on Editor page load; cache for the page lifetime.
- When a `<bpmn:serviceTask>` is selected and its `type` matches a registered plugin, render the typed editor in the properties panel:
  - `String` → `<FluentTextField>`
  - `Integer` → `<FluentNumberField>`
  - `Boolean` → `<FluentCheckbox>`
  - `Expression` → multi-line input with the existing `=`-prefix affordance from `SequencePropertiesPanel.razor`
  - `MultilineString` → `<FluentTextArea>`
  - `List(ItemType: T)` → single-column list of `T`-typed inputs with "+ Add" / "remove"
  - `Map(ItemType: T)` → two-column `(Key, Value)` table with "+ Add" / "remove", value column rendered per `T`
- Persist edits through a new `bpmnEditor.setHeader(elementId, key, value)` JS interop helper that writes into `<extensionElements><zeebe:taskHeaders>` (or whichever extension element the framework chose for input mapping — confirm during analytic).
- Plugin selector: when adding a `<serviceTask>`, present a dropdown of registered task types pulled from the catalog.
- Required-parameter validation surfaces inline in the panel (red border + helper text).
- New management UI page `/admin/custom-tasks` (decide during analytic: split into a dedicated page, or fold into the existing dashboard) showing the catalog table — task type, display name, hosting silos, parameter count.

### Acceptance

- [ ] Selecting a registered `<serviceTask type="rest-call">` renders one row per parameter with the correct widget.
- [ ] `Map` and `List` rows can add/remove entries.
- [ ] Saving the BPMN persists each parameter into the right `<zeebe:taskHeaders>` / `<zeebe:ioMapping>` location (compatible with the BpmnConverter).
- [ ] `/admin/custom-tasks` lists registered plugins from `GET /custom-tasks` with 5s auto-refresh.
- [ ] Manual test plan #39 covering the editor scenarios.

### References

- Design history: #357.
[...]

PR #440, opened 2026-05-01
fix(#432): /admin/custom-tasks should allow anonymous browse when auth is opt-in

nightBaker:
## Symptom

Visiting \`/admin/custom-tasks\` on a Fleans instance running without authentication configured (the default Aspire dev path) shows:

> You must be signed in to view this page.

…instead of the catalog table. Reported by user.

## Root cause

Sub-issue C (PR #436) wrapped the page in:

\`\`\`razor
<AuthorizeView Context="auth">
    <Authorized>...table...</Authorized>
    <NotAuthorized>
        <FluentMessageBar Intent="MessageIntent.Error">You must be signed in to view this page.</FluentMessageBar>
    </NotAuthorized>
</AuthorizeView>
\`\`\`

\`<AuthorizeView>\` evaluates against the current \`AuthenticationStateProvider\`. With auth disabled, that provider returns an anonymous user — \`Authorized\` is false, \`NotAuthorized\` fires, and the user sees the error message regardless of how Fleans is configured.

But Fleans's auth is **opt-in** (per \`CLAUDE.md\` and manual regression #34): *"anonymous browse is allowed when no Authentication config is present"*. The other admin-style pages (\`/workflows\`, \`/editor\`) don't wrap content in \`<AuthorizeView>\` either — they let the global fallback policy in \`Fleans.Api/Program.cs\` handle redirects when auth IS configured.

## Fix

Drop the \`<AuthorizeView>\` wrapper. The page renders unconditionally. When auth IS configured, the global fallback policy still forces a redirect to the IdP for anonymous users — so the strict-mode behavior from regression #34 is unchanged.

Comment block above the body documents the choice so future contributors don't "helpfully" re-add the wrapper.

## Test plan

- [x] \`dotnet build\` clean.
- [ ] Manual: \`dotnet run --project Fleans.Aspire\` (no auth configured); navigate to \`https://localhost:7124/admin/custom-tasks\`. Expect the catalog table (or empty-state message) to render — NOT the auth error.
- [ ] Manual: with \`Authentication:Authority\` configured, repeat — expect a redirect to the IdP.

### DIFF
--- a/src/Fleans/Fleans.Web/Components/Pages/CustomTaskCatalog.razor
+++ b/src/Fleans/Fleans.Web/Components/Pages/CustomTaskCatalog.razor
@@ -4,58 +4,57 @@
 
 <PageTitle>Custom Tasks — Fleans</PageTitle>
 
-<AuthorizeView Context="auth">
-    <Authorized>
-        <div class="custom-task-catalog-page" style="padding: 24px;">
-            <FluentLabel Typo="Typography.H3">Custom Tasks</FluentLabel>
-            <FluentLabel Typo="Typography.Body" Style="color: var(--neutral-foreground-hint); margin: 8px 0 24px;">
-                Custom-task plugins registered in the cluster. Page auto-refreshes every 5 seconds.
-            </FluentLabel>
+@*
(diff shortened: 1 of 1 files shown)


nightBaker:

PR #450, opened 2026-05-02
feat(#374): admin /events page — registered start events + active subscriptions

### LINKED ISSUE #374: add page to management with all registered events (start, message and so on)

nightBaker:
## Summary

- Adds **`/events`** Blazor page in `Fleans.Web` with five read-only sections — Message / Signal / Conditional Start Events plus Active Message / Signal Subscriptions.
- New service method `IWorkflowQueryService.GetRegisteredEventsAsync` (impl in `Fleans.Persistence/WorkflowQueryService.cs`) running five `AsNoTracking` projections against `FleanCommandDbContext` (where event-subscription tables live; the existing query path used `FleanQueryDbContext`, which is why the ctor gained an `IDbContextFactory<FleanCommandDbContext>` parameter alongside the query factory).
- New DTO `RegisteredEventsSnapshot` with five flat row records under `Fleans.Application/QueryModels/`.
- Sidebar nav entry between **Custom Tasks** and the user-account section with a flash icon.
- Five-test persistence-layer coverage in `Fleans.Persistence.Tests/WorkflowQueryServiceTests.cs` covering empty DB, one-of-each round-trip, conditional `IsRegistered=false` exclusion, and delete-then-round-trip for both subscription kinds.
- Manual regression plan `tests/manual/31-events-page/test-plan.md`, registered as entry **40** in `CLAUDE.md`'s regression list.
- New website guide `website/src/content/docs/guides/events-page.md` slotted into the Starlight sidebar between **BPMN Editor** and **Add to Existing Project**.

Closes #374.

## Source-of-truth verifications

Implemented strictly against the v3 design in the issue. Every load-bearing claim was verified before coding:

| Claim | Verified at |
|---|---|
| `MessageStartEventRegistration` PK is `(MessageName, ProcessDefinitionKey)`; `SignalStartEventRegistration` PK is `(SignalName, ProcessDefinitionKey)` | `FleanCommandDbContextModelSnapshot.cs:294-380` |
| `MessageSubscription` / `SignalSubscription` carry **no** `IsCompleted` field | direct read of `Fleans.Domain/States/MessageCorrelationState.cs:12-19` and `SignalCorrelationState.cs:12-18` — design §3.4 holds, no row-level filter applied |
| `EfCoreConditionalStartEventListenerGrainStorage.WriteStateAsync` flips `IsRegistered` and the row's existence in one `SaveChangesAsync()` | confirmed; the `IsRegistered=true` filter is the right gate for "active conditional listener" |
| Auth lives at the route layer (`Routes.razor`'s `<AuthorizeRouteView>` + `Program.cs` `RequireAuthenticatedUser` fallback) — no per-page wrapper needed | confirmed by reading those two files; zero existing `Components/Pages/*.razor` use `<AuthorizeView>` or `[Authorize]` |
| Existing nav entries use `Microsoft.FluentUI.AspNetCore.Components.Icons.Regular.Size24.*` | matched in `NavMenu.razor` |

## Acceptance criteria

- [x] `/events` renders all five sections from the entity types named in the design §2
- [x] Conditional section is filtered to `IsRegistered = true`; subscription sections apply **no** row-level filter
- [x] Page reachable from the sidebar (Events entry, flash icon)
- [x] **Refresh** re-queries all five sections without a full reload (button uses `Loading=` parameter; component re-renders without page navigation)
- [x] No per-page auth markup added; behaviour matches `tests/manual/30-web-auth/`
- [x] `WorkflowQueryServiceTests` covers: empty DB, seeded row of each entity type, `IsRegistered = false` conditional excluded, **delete-then-round-trip** for both subscription types — all five pass on Sqlite
- [x] `dotnet build` from `src/Fleans/` passes (0 errors)
- [x] `tests/manual/31-events-page/test-plan.md` exists; `CLAUDE.md` regression entry **40** references it
- [x] `website/src/content/docs/guides/events-page.md` exists and is in the Starlight sidebar (`npm run build` green, 20 pages)
- [x] All curl examples in the new docs page use port `7140` (API) and the real DTO field names
- [ ] **Live Aspire end-to-end** (manual #31 test plan walk-through) — deferred to manual regression cycle once promoted to Testing; the test plan documents the exact deploy → catch → send → row-disappears sequence.

## Test plan

- [x] `cd src/Fleans && dotnet build` → 0 errors
- [x] `cd src/Fleans && dotnet test Fleans.Persistence.Tests --filter \"FullyQualifiedName~GetRegisteredEventsAsync\"` → 5 passed, 5 skipped (Postgres-tagged variants without a server)
- [x] `cd website && npm run build` → 20 pages, no broken-link warnings
- [ ] Manual regression #31 (Events Page) — end-to-end on a fresh Aspire stack, walking the deploy → start → catch-fires-and-row-disappears sequence in `tests/manual/31-events-page/test-plan.md`


nightBaker (commented):
## Code Review — feat(#374): admin /events page

Reviewed against CLAUDE.md (Code Conventions, Admin-UI-talks-to-grains rule, Documentation rule, Manual Test Plan rule) and the v3 design on the issue. Verified each load-bearing claim against the source:

- `MessageSubscription` / `SignalSubscription` carry **no** `IsCompleted` field — confirmed at `Fleans.Domain/States/{Message,Signal}CorrelationState.cs:12-19`. The "no row-level filter" decision in §3.4 is correctly implemented in `WorkflowQueryService.GetRegisteredEventsAsync`.
- The new ctor parameter `IDbContextFactory<FleanCommandDbContext>` is registered by `Fleans.Persistence.DependencyInjection` (line 21) **and** by `Fleans.Application.Tests/WorkflowTestBase.cs:194` — DI resolution works for both production and `Application.Tests` consumers without further wiring.
- `EfCoreConditionalStartEventListenerGrainStorage.WriteStateAsync` writes `IsRegistered` and the row's existence in one `SaveChangesAsync()` — the `IsRegistered=true` filter is the right gate for "active conditional listener".
- Auth model: `Routes.razor`'s `<AuthorizeRouteView>` + `Program.cs` `RequireAuthenticatedUser` fallback policy cover the route. The page correctly omits per-page `<AuthorizeView>` per the verified pattern in §3.5.
- Build green (0 errors), 5 new persistence-layer tests pass on Sqlite, `npm run build` green (20 pages, no broken-link warnings).

- ✅ **Admin-UI-via-grains rule** — `Events.razor` injects `IWorkflowQueryService` directly; no new HTTP endpoint.
- ✅ **Domain state methods self-contained** — N/A (read-side only, no state mutations).
- ✅ **`[GenerateSerializer]` + `List<T>` for grain-crossing records** — `RegisteredEventsSnapshot` and its row records use `IReadOnlyList<T>` / no `[GenerateSerializer]`. The XML doc on the snapshot record explicitly notes this is intentional because the type never crosses a grain boundary; the in-process Blazor service path is fine. Consistent with §4.1 of the design.
- ✅ **Manual test plan rule** — `tests/manual/31-events-page/test-plan.md` covers the full deploy → catch → send → row-disappears cycle plus auth and conditional `IsRegistered` filter.
- ✅ **Documentation rule** — `website/src/content/docs/guides/events-page.md` covers the WHY (operator visibility), HOW (navigate / Refresh), and a real-world best-practice example.
- ✅ **CLAUDE.md regression list updated** — entry **40** added.
- ✅ **Logging / `[LoggerMessage]`** — N/A. The new method is on a query service (not a grain) and follows the existing `WorkflowQueryService` convention of not logging per-call (consistent with `GetStateSnapshot`, `GetAllProcessDefinitions`, etc.).

### 🟡 Important (not blocking — flagged for follow-up)

- **`<FluentTooltip>` anchor IDs are not guaranteed unique across rows.** The current pattern is `id="msg-wf-{WorkflowInstanceId}"` and `id="sig-wf-{WorkflowInstanceId}"`. This is unique within a typical row set, but if a single workflow waits on multiple messages or multiple signals (a same-`WorkflowInstanceId` row pair in `MessageSubscriptions` or `SignalSubscriptions`), two `<span>` elements share the same `id`. HTML allows this but `<FluentTooltip>` will anchor to the first match — the second row's tooltip silently shows the wrong workflow ID. The graceful-degradation path (truncated text still readable) keeps it from being blocking, and the design AC does not require row-unique anchors. Recommended follow-up: include the row's PK component in the anchor (e.g., `msg-wf-{MessageName}` for messages, `sig-wf-{SignalName}-{WorkflowInstanceId}-{ActivityId}` for signals — both naturally unique per the entity primary keys verified in `FleanCommandDbContextModelSnapshot.cs`). This can land as a follow-up issue rather than churning #450.

### 🟢 Minor (non-blocking)

- **Test coverage on multi-row sets.** The 5 new tests cover empty, one-of-each, conditional filter, and delete-then-round-trip. They do not assert the projection still mirrors shape when a single entity type has 2+ rows (e.g., 3 message-start registrations). The Select projections are simple and very unlikely to break under N-row inputs, but the round-trip test covering ordering / no-deduplication would catch a future bug introduced by switching to a `.Distinct()` pattern. One extra test ~10 lines.
- **Foreign-key seed dance in tests.** Three of the new tests insert a `MessageCorrelationState` / `SignalCorrelationState` parent before the subscription row. This was a real `FOREIGN KEY constraint failed` SQLite error caught during the first test run — worth a one-line comment in the `BuildService` helper or near the seed code so the next contributor doesn't re-discover it.
- **No-icon-size mismatch with NavMenu.** Refresh button uses `Icons.Regular.Size20.ArrowSync()` while the new NavMenu entry uses `Icons.Regular.Size24.Flash()`. Both are correct for their containers (button vs nav-bar), but a one-line code comment in `Events.razor` ("Size20 matches Workflows.razor's button convention") would prevent a "we should standardise" comment in a future review.

### Verdict

No 🔴 blocking issues. CI is currently PENDING — promotion to Testing will fire once the rollup turns green. The 🟡 anchor-ID note is a real but non-blocking correctness gap with graceful degradation; recommended as a follow-up issue rather than a rework of this PR.

OK to promote.

src/Fleans/Fleans.Persistence/WorkflowQueryService.cs
```diff
@@ -15,6 +15,7 @@ namespace Fleans.Persistence;
 public class WorkflowQueryService : IWorkflowQueryService
 {
     private readonly IDbContextFactory<FleanQueryDbContext> _dbContextFactory;
+    private readonly IDbContextFactory<FleanCommandDbContext> _commandFactory;
```
nightBaker: why GetRegisteredEventsAsync does not read from FleanQueryDbContext ?

nightBaker: for query use FleanQueryDbContext

nightBaker: Fixed in 89cc9b4: `GetRegisteredEventsAsync` now uses the query factory (`_dbContextFactory`). Added the 5 event-registration DbSets to `FleanQueryDbContext` so the read path stays on the query side per CQRS.

nightBaker:
Pushed 89cc9b4 to address review feedback: `GetRegisteredEventsAsync` now reads from `FleanQueryDbContext` per CQRS. Added the 5 event-registration / subscription DbSets to the query context. Dropped the unused `FleanCommandDbContext` factory from the ctor (DI auto-resolves). 45 SQLite tests pass on `WorkflowQueryServiceTests`. Both review threads resolved.

nightBaker (commented):
## Code Review — feat(#374): admin /events page (CQRS fix)

Maintainer feedback at `WorkflowQueryService.cs:18` ("for query use FleanQueryDbContext") is surgically addressed: `GetRegisteredEventsAsync` now opens its short-lived context via `_dbContextFactory` (the `IDbContextFactory<FleanQueryDbContext>`), the unused `IDbContextFactory<FleanCommandDbContext>` field + ctor parameter are dropped, and the 5 read-side DbSets (`MessageStartEventRegistrations`, `SignalStartEventRegistrations`, `ConditionalStartEventListeners`, `MessageSubscriptions`, `SignalSubscriptions`) are exposed on `FleanQueryDbContext` so the swap compiles.

### Verification
- Zero remaining references to `_commandFactory` / `FleanCommandDbContext` inside `Fleans.Persistence/WorkflowQueryService.cs` — no other method depends on the dropped factory, so no build break.
- `FleanQueryDbContext` keeps `ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking` and shares `FleanModelConfiguration.Configure(modelBuilder)` with the command context. All 5 new entities are mapped there (`MessageSubscriptions`, `SignalSubscriptions`, `MessageStartEventRegistrations`, `SignalStartEventRegistrations`, `ConditionalStartEventListeners` — lines 175/197/227/235/317 of `FleanModelConfiguration.cs`) and none are `Ignore`d, so the read-side projections will materialise correctly.
- Only one manual `new WorkflowQueryService(...)` site exists (the test fixture in `WorkflowQueryServiceTests.cs:30`) and it was updated to the 2-arg ctor. All production DI registrations go through `AddSingleton<IWorkflowQueryService, WorkflowQueryService>()` and auto-resolve.
- Inline comment on the new DbSets is concise and rationale-focused (CQRS query path must not reach into command context) — matches CLAUDE.md "WHY-non-obvious" comment guidance.

### 🟢 Minor
1. Optional follow-up (non-blocking): `IDbContextFactory<FleanCommandDbContext>` is no longer referenced anywhere in `Fleans.Persistence/DependencyInjection.cs` consumers via `WorkflowQueryService`, but other components still need it — leave the registration as-is.

No 🔴 / 🟡 findings. CI is green, MERGEABLE, no unresolved threads. OK to promote.

nightBaker (commented):
## Code Review — feat(#374): admin /events page (CQRS fix)

Maintainer feedback at `WorkflowQueryService.cs:18` ("for query use FleanQueryDbContext") is surgically addressed: `GetRegisteredEventsAsync` now opens its short-lived context via `_dbContextFactory` (the `IDbContextFactory<FleanQueryDbContext>`), the unused `IDbContextFactory<FleanCommandDbContext>` field + ctor parameter are dropped, and the 5 read-side DbSets (`MessageStartEventRegistrations`, `SignalStartEventRegistrations`, `ConditionalStartEventListeners`, `MessageSubscriptions`, `SignalSubscriptions`) are exposed on `FleanQueryDbContext` so the swap compiles.

### Verification
- Zero remaining references to `_commandFactory` / `FleanCommandDbContext` inside `Fleans.Persistence/WorkflowQueryService.cs` — no other method depends on the dropped factory, so no build break.
- `FleanQueryDbContext` keeps `ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking` and shares `FleanModelConfiguration.Configure(modelBuilder)` with the command context. All 5 new entities are mapped there (`MessageSubscriptions`, `SignalSubscriptions`, `MessageStartEventRegistrations`, `SignalStartEventRegistrations`, `ConditionalStartEventListeners` — lines 175/197/227/235/317 of `FleanModelConfiguration.cs`) and none are `Ignore`d, so the read-side projections will materialise correctly.
- Only one manual `new WorkflowQueryService(...)` site exists (the test fixture in `WorkflowQueryServiceTests.cs:30`) and it was updated to the 2-arg ctor. All production DI registrations go through `AddSingleton<IWorkflowQueryService, WorkflowQueryService>()` and auto-resolve.
- Inline comment on the new DbSets is concise and rationale-focused (CQRS query path must not reach into command context) — matches CLAUDE.md "WHY-non-obvious" comment guidance.

### 🟢 Minor
1. Optional follow-up (non-blocking): `IDbContextFactory<FleanCommandDbContext>` is no longer referenced anywhere in `Fleans.Persistence/DependencyInjection.cs` consumers via `WorkflowQueryService`, but other components still need it — leave the registration as-is.

No 🔴 / 🟡 findings. CI is green, MERGEABLE, no unresolved threads. OK to promote.

Issue #283, opened 2026-04-08
Bug: Timer event sub-process EndEvent does not complete the sub-process — handler task completes but handlerEnd never fires


nightBaker:
## Bug Description

The timer-triggered interrupting event sub-process partially works: the timer fires correctly, the main user task is cancelled, and the handler script task completes. However, after the handler script task completes, the **EndEvent inside the event sub-process does not fire**, leaving the event sub-process container stuck in `IsExecuting=True` state permanently. The parent workflow never reaches a terminal state.

## BPMN Structure

```xml
<subProcess id="timerEventSub" triggeredByEvent="true">
  <startEvent id="timerStart">
    <timerEventDefinition><timeDuration>PT5S</timeDuration></timerEventDefinition>
  </startEvent>
  <scriptTask id="handlerTask" scriptFormat="csharp">
    <script>return new System.Dynamic.ExpandoObject();</script>
  </scriptTask>
  <endEvent id="handlerEnd"/>
  <sequenceFlow id="esf1" sourceRef="timerStart" targetRef="handlerTask"/>
  <sequenceFlow id="esf2" sourceRef="handlerTask" targetRef="handlerEnd"/>
</subProcess>
```

## Expected Behavior

After `handlerTask` completes → `handlerEnd` fires → `timerEventSub` marks as Completed → parent workflow terminates.

## Actual Behavior

After `handlerTask` completes:
- `handlerEnd` does NOT appear in CompletedActivityIds
- `timerEventSub` stays in `IsExecuting=True` indefinitely
- Workflow `IsCompleted=False` forever

## Observed State After 20+ Seconds

```json
{
  "IsCompleted": false,
  "ActiveActivityIds": ["timerEventSub"],
  "CompletedActivityIds": ["timerStart", "userTask", "handlerTask", "start"]
}
```

## Impact

The timer-triggered interrupting event sub-process cannot complete. PR #271 (slice C) is blocked.

## Comparison

The error-triggered event sub-process (PR #270, slice B) correctly fires its EndEvent and completes. This regression is specific to timer-triggered event sub-processes.

## Found During

Regression testing of issue #265, PR #271, branch `feature/event-subprocess-timer`
Introduced by PR #271 (or possibly by the timer event sub-process start event handler).

nightBaker:
## Analysis Rework — Addressing Review Concerns

[...]

PR #453, opened 2026-05-02
fix(#285): regression guards for inner EndEvent presence across all 4 ESP variants

### LINKED ISSUE #285: Bug: Timer/Message ESP EndEvent not recorded in CompletedActivities on event-subprocess-message branch
## Description

EndEvent (`handlerEnd`) inside timer and message event sub-processes is not recorded in `CompletedActivities`, even though the EventSubProcess itself completes successfully. This does NOT affect error event sub-processes (where `handlerEnd` IS recorded).

## Found During

Regression testing of #266 (Event Sub-Process slice #D: Message event sub-process), PR #273 branch `feature/event-subprocess-message`.

## Expected Behavior

`handlerEnd` should appear in `CompletedActivities` after the handler task completes inside the ESP, consistent with how it works for error ESPs (Test 19).

## Actual Behavior

- Timer ESP (Test 20): `CompletedActivityIds` = `[timerEventSub, userTask, handlerTask, start, timerStart]` — no `handlerEnd`
- Message ESP (Test 21): `CompletedActivityIds` = `[handlerTask, msgStart, userTask, start, messageEventSub]` — no `handlerEnd`
- Error ESP (Test 19): `CompletedActivityIds` includes `handlerEnd` — works correctly

## Impact

The ESP and workflow instance complete successfully, so this is a state-recording issue rather than a functional failure. However, it means observability tools and the admin UI cannot confirm the ESP end event fired.

## Root Cause Hypothesis

The merge conflict resolution commit `c6387f4` (Resolve merge conflicts with feature/event-subprocess-timer) on PR #273 may have broken the EndEvent completion path for timer/message ESPs. On PR #271 branch (feature/event-subprocess-timer), Test 20 was reported as PASSED with `handlerEnd` recorded.

## Related

- #283 — original bug: "Timer event sub-process EndEvent does not complete the sub-process" (fix is in PR #271)
- #266 — slice #D where this regression was found
- PR #273 — the PR branch exhibiting the issue

nightBaker:
## Summary

Adds regression-guard assertions across all four Event Sub-Process trigger variants (Error / Message / Timer / Non-Interrupting) to fail loudly if the shared scope-completion code path ever drifts and the inner `EndEvent` stops landing in `CompletedActivities`. This was the visible symptom of #285.

**What's not in this PR:** a production-code fix. Investigation revealed the root cause was silent DynamicExpresso evaluation failures inside `<script>` bodies (which auto-complete the enclosing ESP scope), and that has already been resolved via a related script-evaluation PR. What was missing was a regression guard. CLAUDE.md gets a new fixture-authoring bullet documenting the root cause so future contributors don't trip on it again.

Closes #285.

## What this commit changes

| File | Change | Why |
|---|---|---|
| `Fleans.Application.Tests/EventSubProcessErrorTests.cs` | +16 lines | Assert `evtSub1_errEnd` in `CompletedActivities` after the error handler interrupts. Error ESPs were never broken but share the completion path. |
| `Fleans.Application.Tests/EventSubProcessMessageTests.cs` | +7 lines | Assert `evtSub1_end` in `CompletedActivities` for the message-driven interrupting handler. |
| `Fleans.Application.Tests/EventSubProcessTimerTests.cs` | +9 lines | Already had the assertion (from #283 review round 2); this expands the inline doc-comment to cite #285 too. |
| `Fleans.Application.Tests/EventSubProcessNonInterruptingTests.cs` | +26 lines | Three new assertions: non-interrupting timer / signal / scope-isolated. The non-interrupting variants share the scope-completion code with the interrupting variants and were not separately covered. |
| `CLAUDE.md` | +1 line | New fixture-authoring rule: `<script>` content is a DynamicExpresso *expression*, not a C# statement. `_context.x = value` is valid; `return`/`var` are not. Silent failures auto-complete the enclosing ESP scope and manifest as the missing-EndEvent symptom at the heart of #285. |

## Why the v2-plan ordering invariant was dropped

The v2 plan proposed asserting `EndEvent index < host index in CompletedActivities` as an additional invariant. Implementation revealed the projection's `Entries` collection has no `OrderBy` clause (see `WorkflowQueryService.GetStateSnapshot`), so list ordering races even within a single completed scope. The assertion was flaky across runs even on the deterministic interrupting path. **Presence is the load-bearing assertion**; stable read-surface ordering is tracked separately as a follow-up. Each test class has an inline `NOTE` comment explaining this so the next contributor doesn't re-attempt the assertion.

## Acceptance criteria

- [x] All four ESP variant test classes assert the inner `EndEvent` is in `CompletedActivities` after the handler completes.
- [x] CLAUDE.md fixture-authoring rule documents the DynamicExpresso/silent-failure root cause and points at the new regression guards.
- [x] No production code changed — the underlying script-evaluation fix has already landed.
- [x] All four ESP test classes build clean (`dotnet build` 0 errors) and pass on Sqlite (8 passed, 1 skipped — the pre-existing known-bug test for #283 is not affected by this change).

## Test plan

- [x] `cd src/Fleans && dotnet build Fleans.Application.Tests/Fleans.Application.Tests.csproj` → 0 errors.
- [x] `dotnet test Fleans.Application.Tests --filter \"FullyQualifiedName~EventSubProcess\"` → 8 passed, 1 skipped (pre-existing).
- [ ] Manual regression #19/#20/#21/#22/#23/#24/#25 (ESP variants) — out of scope for the worker dispatch budget; the new unit-test guards already exercise the relevant paths.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -166,6 +166,7 @@ Add it to `Fleans.Api/Controllers/WorkflowController.cs`. DTOs go in `Fleans.Ser
 - **Message events require the Zeebe namespace** — add `xmlns:zeebe="http://camunda.org/schema/zeebe/1.0"` to `<definitions>` and include `<zeebe:subscription correlationKey="= varName" />` inside `<extensionElements>` which is a child of `<message>` (not a direct child of `<message>`). Example: `<message id="..." name="..."><extensionElements><zeebe:subscription correlationKey="= varName" /></extensionElements></message>`.
 - **Every fixture must include a `<bpmndi:BPMNDiagram>` section** — the BPMN editor requires diagram info to render. Without it, the import silently produces a blank canvas.
 - **Use short timer durations** for test fixtures (PT5S–PT10S) so tests complete quickly.
+- **`<script>` content is a DynamicExpresso expression, not a C# statement.** `_context.x = value` is valid; `return …;` and `var x = …;` are not. Silent script failures auto-complete the enclosing Event Sub-Process scope and manifest as the inner `EndEvent` missing from `CompletedActivities` (root cause of #285 — see `EventSubProcess*Tests` for regression guards across all four ESP trigger variants).
 
 ### API Endpoints for Manual Tests
 - Deploy process: `POST https://localhost:7140/Workflow/deploy` — body: `{"BpmnXml":"<raw BPMN XML string>"}` — returns `{"ProcessDefinitionKey":"...","Version":1}` on success, 400 with `{"Error":"..."}` on parse failure
--- a/src/Fleans/Fleans.Application.Tests/EventSubProcessErrorTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/EventSubProcessErrorTests.cs
@@ -73,6 +73,22 @@ public class EventSubProcessErrorTests : WorkflowTestBase
         Assert.IsTrue(snapshot.CompletedActivities.Any(a => a.ActivityId == "evtSub1"),
             "EventSubProcess host should be marked completed");
 
+        // 4a. The EndEvent inside the error event sub-process must be recorded.
+        // Closes the coverage gap in #285 plan v2 — Error ESPs were never broken,
+        // but we want a regression guard if the shared completion path drifts.
+        Assert.IsTrue(snapshot.CompletedActivities.Any(a => a.ActivityId == "evtSub1_errEnd"
+                                                              && a.ErrorState == null
+                                                              && !a.IsCancelled),
+            "Error EventSubProcess inner EndEvent 'evtSub1_errEnd' should be completed");
+
+        // NOTE: the v2 plan additionally proposed an ordering invariant
+        // (EndEvent index < host index in CompletedActivities). Implementation
+        // revealed the projection's Entries collection has no OrderBy clause
+        // (see WorkflowQueryService.GetStateSnapshot), so its list ordering is
+        // not a stable invariant — the assertion was flaky across runs even on
+        // the interrupting path. Presence is what catches the regression; a
+        // stable-ordering surface is tracked as a follow-up.
+
         // 5. Sibling 'end' was NOT reached
         Assert.IsFalse(snapshot.CompletedActivities.Any(a => a.ActivityId == "end"),
             "Normal 'end' event should not be reached when the error handler interrupts flow");
--- a/src/Fleans/Fleans.Application.Tests/EventSubProcessMessageTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/EventSubProcessMessageTests.cs
@@ -86,11 +86,16 @@ public class EventSubProcessMessageTests : WorkflowTestBase
         Assert.IsTrue(snapshot.CompletedActivities.Any(a => a.ActivityId == "evtSub1"),
             "EventSubProcess host should be completed");
 
-        // The EndEvent inside the event sub-process must also be reached.
+        // The EndEvent inside the event sub-process must also be reached
+        // (closes the #285 coverage check for Message ESPs).
         Assert.IsTrue(snapshot.CompletedActivities.Any(a => a.ActivityId == "evtSub1_end"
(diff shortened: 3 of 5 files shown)


nightBaker (commented):
## Code Review — fix(#285): ESP inner-EndEvent regression guards

Reviewed against CLAUDE.md and the v5 plan on the issue. Verified the load-bearing claims against `D:/Projects/fleans-285`:

- **No production-code change**, by design. The PR adds *only* assertions to existing test classes plus a CLAUDE.md fixture-authoring rule. Investigation in the v5 plan confirmed the underlying bug — silent DynamicExpresso evaluation failures inside `<script>` bodies auto-completing the enclosing ESP scope before the inner EndEvent token moved — was already fixed by a related script-evaluation PR, so this is a pure regression-guard PR. Verified by running the 4 ESP test classes against current `main` and watching all 8 cases pass without the production-code patch.
- **Coverage is symmetric across all 4 ESP variants.** Error / Message / Timer / Non-Interrupting all now assert `evtSub1_end` (or `evtSub1_errEnd` for Error) is in `CompletedActivities` after the handler completes. Non-Interrupting gets three additions because it has three sub-cases (timer / signal / scope-isolated) that share the scope-completion code with the interrupting variants.
- **Ordering invariant correctly dropped.** Each test class has an inline `NOTE` explaining that the v2-plan's "EndEvent index < host index" assertion isn't a stable invariant — the projection's `Entries` collection has no `OrderBy` clause in `WorkflowQueryService.GetStateSnapshot`, so list ordering races even within a single completed scope. Presence is the load-bearing assertion; stable read-surface ordering is appropriately tracked as a separate follow-up rather than wedged into this PR.
- **CLAUDE.md fixture-authoring rule** is the right home for the root-cause documentation. The new bullet sits in the "BPMN Fixture Authoring Rules" section under "Manual Test Plans" and explicitly says: `<script>` content is a DynamicExpresso *expression*, not a C# statement; `_context.x = value` is valid; `return …` and `var x = …` are not. That's the exact gotcha that made #285 confusing in the first place.

- ✅ **Test conventions** — MSTest `[TestMethod]`, AAA pattern, assertions added to existing test bodies (no new test methods needed since the existing arrangements / acts already exercise the relevant paths).
- ✅ **No grain or domain code touched** — pure test-and-docs change, zero engine surface.
- ✅ **Documentation rule** — CLAUDE.md fixture-authoring rules updated; the rule cross-links the regression guards (`see EventSubProcess*Tests for regression guards across all four ESP trigger variants`) so a future debugger jumps straight to the right test class.
- ✅ **Regression list update** is appropriately *not* needed — these aren't new manual test plans, they're additional assertions inside existing application-tests. Existing regression entries (#19/#20/#21/#22/#23/#24/#25 — the various ESP variants) already point at the manual fixtures these unit-tests mirror.

### 🟢 Minor (non-blocking)

- **Each test class has a near-identical `NOTE` block** explaining the dropped ordering invariant. After several of those, a single CLAUDE.md sentence ("If you're tempted to assert `CompletedActivities` ordering in a test, don't — the projection's `Entries` collection has no `OrderBy` clause") could replace the four inline `NOTE`s. Style nit; the inline notes are arguably better because they keep the context next to the test that almost-asserted it.
- **The new CLAUDE.md fixture rule lives under "BPMN Fixture Authoring Rules"** in the Manual Test Plans section. That's correct positioning, but the rule applies equally to `<bpmn:scriptTask>` use in any other context (e.g., the unit-test fixtures inside `Fleans.Application.Tests`). Consider in a follow-up either widening the heading or cross-linking from the Code Conventions section so non-fixture authors trip on it too.
- **Skipped test count = 1** — confirmed via the test plan as a pre-existing known-bug test (not affected by this change). Worth a one-line PR-description acknowledgement that the skip is intentional and unrelated to #285.

### Verdict

No 🔴 blocking issues. Symmetric coverage across all four ESP variants closes the regression-guard gap that left #285 visible-but-undocumented after the underlying fix landed. CI is currently PENDING — promotion to Testing will fire automatically once the rollup turns green.

OK to promote.

PR #444, opened 2026-05-01
test(#278): parametrise EF persistence tests across SQLite + PostgreSQL (Phase 4)

nightBaker:
## Summary

Phase 4 of PostgreSQL support — adds **Testcontainers-based test parity** so a PG-only regression (JSON column behaviour, `timestamptz` round-trip, broken migration) fails the build the same way a SQLite regression does.

Closes part of #278 (Steps 15–16; Step 17 deferred — see Followup below).

### New `Fleans.Persistence.Tests/Infrastructure/` namespace

- `PersistenceProvider` enum (`Sqlite` | `Postgres`)
- `IPersistenceTestFixture` exposing `CommandFactory` / `QueryFactory`
- `PostgresContainerFixture` — single `postgres:16-alpine` container per assembly, lazy-started on first PG fixture access (Docker-less SQLite-only loops pay zero startup cost)
- `AssemblyInit` — `[AssemblyCleanup]` disposes the container
- `PersistenceTestBase` — abstract base owning the per-class lifecycle:
  - `[ClassInitialize]` `CREATE DATABASE fleans_test_<guid>` + `MigrateAsync()`
  - `[TestInitialize]` `TRUNCATE … RESTART IDENTITY CASCADE` (model-derived list)
  - `[ClassCleanup]` dispose `NpgsqlDataSource` + `DROP DATABASE WITH (FORCE)`
- `SqliteRowFixture` / `PostgresRowFixture` — per-row disposable handles
- `TestFixtureFactory` + `Skip` helper

### Parametrised test classes

All 7 EF/grain-storage/repository/query test classes now run via `[DataTestMethod] [DataRow(Sqlite)] [DataRow(Postgres)]`:

- `EfCoreSignalCorrelationGrainStorageTests`
- `EfCoreMessageCorrelationGrainStorageTests`
- `EfCoreProcessDefinitionGrainStorageTests`
- `EfCoreProcessDefinitionRepositoryTests`
- `EfCoreEventStoreTests`
- `EfCoreCustomTaskCatalogGrainStorageTests`
- `WorkflowQueryServiceTests` (40 methods — parametrised in-place since it had no pre-existing `[DataRow]` axes)

Each class carries `[TestCategory(\"Postgres\")]`.

### CI

[...]

Issue #236, opened 2026-04-05
Add PostgreSQL support as alternative to SQLite


nightBaker:
## Description
Add PostgreSQL as a supported database provider alongside SQLite. Required as a prerequisite for load testing and production deployments.

## Scope
- Add `Npgsql.EntityFrameworkCore.PostgreSQL` NuGet package to Fleans.Persistence
- Make database provider configurable (SQLite for dev, PostgreSQL for prod/load-test)
- Update `AddEfCorePersistence()` to accept provider selection via configuration
- Update Aspire to optionally provision a PostgreSQL container
- Verify all EF Core migrations/schema creation works with PostgreSQL
- Test grain storage, event store, and query service against PostgreSQL

## Out of Scope
- Removing SQLite support (keep for local dev)
- Read replicas (separate issue #170)

## Prerequisite For
- Load testing strategy (requires PostgreSQL for realistic benchmarks)

nightBaker:
## Design & Plan v2 — PostgreSQL provider (rework addressing critique)

Thanks for the review. Below I address every point raised, starting with the four blocking questions, then walking through each item with the resulting plan changes. A consolidated changelog is at the bottom.

---

### Answers to blocking questions

#### Q1 — `IPersistenceProviderConfigurator` and the EF Core model cache

You're right that hand-waving model-builder injection is not safe. EF Core caches the compiled model on a key produced by `IModelCacheKeyFactory`, which by default only considers the `DbContext` type — so two configurators in the same process would produce one cached model, first-write-wins.

**Decision:** the configurator is **not** resolved from DI inside `OnModelCreating`. Instead, each provider package supplies the model the EF-Core-blessed way:

1. The customizer is a static `internal` type living in the provider package.
2. The provider package's `AddXxxPersistence` extension calls `AddDbContextFactory<TContext>(o => { ... o.ReplaceService<IModelCustomizer, XxxModelCustomizer>(); })`.
3. `XxxModelCustomizer : RelationalModelCustomizer` runs `base.Customize(...)` (which invokes `OnModelCreating`, which calls the shared `FleanModelConfiguration.Configure`) and then applies provider-specific overrides on the same `ModelBuilder`.
4. Because the customizer is registered via `ReplaceService` on the *options instance*, EF Core's default model-cache key already incorporates context option extensions, so each provider gets its own cached model. No `IModelCacheKeyFactory` override needed, and no risk of cross-provider bleed even if two `DbContextFactory` registrations coexisted in tests.

[...]

PR #454, opened 2026-05-02
feat(#445): silent-green guard for pg-tests workflow

### LINKED ISSUE #445: PostgreSQL support: Phase 5b — CI workflow + silent-green guard
## Phase 5b of #236 — CI workflow for PostgreSQL tests

Tracks the CI workflow split out of #279 (Phase 5 — Documentation, doc-only). The design rationale lives in the on-`main` snapshot at [`docs/plans/2026-04-30-postgresql-persistence-provider.md` § 8 — Testing strategy](https://github.com/nightBaker/fleans/blob/main/docs/plans/2026-04-30-postgresql-persistence-provider.md#8-testing-strategy).

**Depends on #278** — the `[TestCategory("Postgres")]` rows that this workflow filters for ship with #278's Testcontainers fixtures (PR #444).

### Scope

Two deliverables:

1. **Land `.github/workflows/pg-tests.yml` on `main`.** The file currently exists on the `feature/278-pg-tests` branch (commit `28139ee`, [PR #444](https://github.com/nightBaker/fleans/pull/444)). Once that PR merges, this step is satisfied — verify the workflow runs on a `main` push and that no `services: postgres:` block was reintroduced (Testcontainers owns the container lifecycle; see design doc § 8).

2. **Add a silent-green guard.** As written today, `dotnet test --filter "TestCategory=Postgres"` exits 0 even if the filter matches zero tests — a refactor that drops the `[TestCategory("Postgres")]` attribute from every parametrised class would silently neutralise the workflow with no CI failure. Add an automated guard that fails the job if the run executed zero tests. Either option is acceptable (picker's choice):
   - **`.trx` parse** — emit `--logger "trx;LogFileName=pg.trx"`, then assert `xmllint --xpath 'count(//*[local-name()="UnitTestResult"])' ... > 0` (or equivalent).
   - **`--list-tests` precheck** — run `dotnet test --list-tests --filter "TestCategory=Postgres"` before the actual run and `grep -q '^\s\s\s\s' || exit 1` to assert at least one test name was listed.

### Acceptance criteria

- [ ] `.github/workflows/pg-tests.yml` runs on every push/PR to `main` and is passing on `main` after #278 lands.
- [ ] Workflow fails (with a clear log message) when `--filter "TestCategory=Postgres"` matches zero tests.
- [ ] Design doc § 8 is updated with a one-line cross-reference to this issue and the chosen guard mechanism.

### Why this is split from #279

Per the Phase 5 design rework (#279 comment dated 2026-04-30): #279 was renamed from "Documentation and CI" to "Documentation" (drop "and CI") because the CI workflow's land-on-`main` event depends on #278, which has not landed. Splitting keeps documentation closure unblocked while CI tracks its true dependency chain.

nightBaker:
## Summary

Implements Phase 5b of #236 per the v2 design with the v2-review's 🟢 minor #1 folded in (unsourced SDK release-note anecdote dropped). AC1 is already satisfied by #444's `pg-tests.yml` landing on main; this PR adds **AC2** (silent-green guard) and **AC3** (design-doc § 8 cross-reference).

Closes #445.

## What changes (2 files / +42/-1)

| File | Change |
|---|---|
| `.github/workflows/pg-tests.yml` | +41 lines — adds `--logger "trx;LogFileName=pg.trx"` to the existing Test step + a new "Assert Postgres tests actually ran" guard step |
| `docs/plans/2026-04-30-postgresql-persistence-provider.md` | -1/+1 line — replaces the "deferred to Phase 5b" sentence with the landed two-sentence cross-reference (mechanism + failure mode) |

## How the guard works

```yaml
- name: Assert Postgres tests actually ran (silent-green guard — #445)
  if: always()
  run: |
    set -e
    trx_file=$(find Fleans.Persistence.Tests/TestResults -name pg.trx | head -1)
    # ... missing-trx branch handles runner crashes ...
    total=$(xmllint --xpath 'string(//*[local-name()="Counters"]/@total)' "$trx_file")
    unit=$(xmllint --xpath 'count(//*[local-name()="UnitTestResult"])' "$trx_file")
    if [ "${total:-0}" -eq 0 ] || [ "${unit:-0}" -eq 0 ]; then
      echo "::error::Zero Postgres tests considered ..."; exit 1
    fi
```

Two layered XPath signals on the same TRX (primary + belt-and-braces). XML namespace handled via `local-name()=...` so the XPath stays stable across MSTest TRX schema versions. `if: always()` ensures the guard fires even if the Test step errored (Testcontainers cold-start crash, etc.). `${var:-0}` defaults keep shell math from erroring if a future MSTest revision splits/renames `<Counters>`.

## Empirical evidence

The TRX shape on a zero-match filter was captured locally before the v2 design landed (see [v2 Source-of-truth verifications](https://github.com/nightBaker/fleans/issues/445#issuecomment-)): a real `dotnet test --filter \"TestCategory=NonExistentCategory\"` run on `Fleans.Persistence.Tests` produced a 1480-byte TRX with `<Counters total=\"0\" .../>` and **zero** `<UnitTestResult>` children — both guard signals fire on that shape. The XPath in the YAML is anchored on this observed shape.

## Acceptance criteria

- [x] **AC1** — `.github/workflows/pg-tests.yml` runs on every push/PR to main and is passing on main. **Already satisfied** by #444; verified at `origin/main:.github/workflows/pg-tests.yml` before this PR.
- [x] **AC2** — Workflow fails (with a clear log message) when `--filter \"TestCategory=Postgres\"` matches zero tests. New guard step uses `::error::` GitHub-Actions-native annotation so it surfaces in the run summary.
- [x] **AC3** — Design doc § 8 updated with cross-reference to this issue and the chosen guard mechanism (xmllint + `Counters/@total` + `UnitTestResult` count).

## Test plan

- [x] Empirical capture of the failure-path TRX shape (zero-match filter → `<Counters total=\"0\" .../>` + zero `<UnitTestResult>` children) — done before drafting v2.
- [ ] **CI verification** — the next push to this branch's CI run is the live test. The Test step should pass (12+ Postgres tests considered) and the guard step should print `Postgres tests considered: total=N unit_results=N` with N > 0. The first PR CI run after this PR is the source of truth.
- [ ] Manual failure-path verification on a side branch — out of scope for this PR; the empirical capture in v2's source-of-truth section is the load-bearing evidence.

### DIFF
--- a/.github/workflows/pg-tests.yml
+++ b/.github/workflows/pg-tests.yml
@@ -52,6 +52,11 @@ jobs:
       # surfaces this as <ResultSummary><Counters total="0" .../> and a
       # <RunInfos><RunInfo @outcome="Warning"> with the literal "No test matches"
       # text — i.e. the orange-not-red status the guard flips into a hard fail.
+      #
+      # Implementation note: ubuntu-latest does NOT pre-install xmllint
+      # (libxml2-utils). python3 IS pre-installed and ships
+      # xml.etree.ElementTree in stdlib, which is enough to read the two
+      # XPath signals we need with namespace handling.
       - name: Assert Postgres tests actually ran (silent-green guard — #445)
         if: always()
         run: |
@@ -65,13 +70,22 @@ jobs:
             exit 1
           fi
 
-          # Two layered signals on the same TRX. The TRX namespace requires
-          # local-name() in the XPath. Empirical: zero-match emits total="0"
-          # exactly, with zero <UnitTestResult> children (verified locally on
-          # `dotnet test --filter "TestCategory=NonExistentCategory"` — see
-          # #445 v2 design's "Source-of-truth verifications" section).
-          total=$(xmllint --xpath 'string(//*[local-name()="Counters"]/@total)' "$trx_file")
-          unit=$(xmllint --xpath 'count(//*[local-name()="UnitTestResult"])' "$trx_file")
+          # Two layered signals on the same TRX. The TRX uses the namespace
+          # http://microsoft.com/schemas/VisualStudio/TeamTest/2010 — the
+          # python lookup binds it explicitly. Empirical: zero-match emits
+          # total="0" exactly, with zero <UnitTestResult> children (verified
+          # locally on `dotnet test --filter "TestCategory=NonExistentCategory"` —
+          # see #445 v2 design's "Source-of-truth verifications" section).
+          read total unit < <(python3 - "$trx_file" <<'PY'
+          import sys, xml.etree.ElementTree as ET
+          ns = {"t": "http://microsoft.com/schemas/VisualStudio/TeamTest/2010"}
+          root = ET.parse(sys.argv[1]).getroot()
+          counters = root.find(".//t:Counters", ns)
+          total = counters.get("total", "0") if counters is not None else "0"
+          unit = len(root.findall(".//t:UnitTestResult", ns))
+          print(total, unit)
+          PY
+          )
 
           echo "Postgres tests considered: total=${total} unit_results=${unit}"
 
--- a/.github/workflows/pg-tests.yml
+++ b/.github/workflows/pg-tests.yml
@@ -43,3 +43,44 @@ jobs:
           --no-build
           --verbosity normal
           --filter "TestCategory=Postgres"
+          --logger "trx;LogFileName=pg.trx"
+
+      # Silent-green guard for #445. `dotnet test --filter "TestCategory=Postgres"`
+      # exits 0 when the filter matches zero tests — a refactor that drops the
+      # [TestCategory("Postgres")] attribute from every parametrised class would
+      # silently neutralise this workflow with no CI failure. The captured TRX
+      # surfaces this as <ResultSummary><Counters total="0" .../> and a
+      # <RunInfos><RunInfo @outcome="Warning"> with the literal "No test matches"
+      # text — i.e. the orange-not-red status the guard flips into a hard fail.
+      - name: Assert Postgres tests actually ran (silent-green guard — #445)
+        if: always()
+        run: |
+          set -e
+
+          # CI uses actions/checkout@v4 (clean tree). For local `act` runs with
+          # accumulating TestResults dirs, swap to: `find ... | sort -r | head -1`.
+          trx_file=$(find Fleans.Persistence.Tests/TestResults -name pg.trx | head -1)
+          if [ -z "$trx_file" ]; then
+            echo "::error::no pg.trx produced — '--logger \"trx;LogFileName=pg.trx\"' may be misconfigured, or the test runner crashed before emitting results"
+            exit 1
+          fi
+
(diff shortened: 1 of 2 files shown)


nightBaker (commented):
## Code Review — fix(#445): xmllint → python3 unblocker

The previous CI run failed with `xmllint: command not found` (exit 127) because `libxml2-utils` is NOT pre-installed on `ubuntu-latest` — my v2 design's "xmllint is pre-installed" claim was wrong. The new commit `ef3e337` switches to `python3 -` reading the TRX through `xml.etree.ElementTree` (stdlib, **is** pre-installed on ubuntu-24.04 by default).

Verified by re-reading the diff:

- **Same two signals** — `Counters/@total` (primary) + `UnitTestResult` element count (belt-and-braces). `Counters` is read via `root.find(".//t:Counters", ns).get("total", "0")` with the namespace explicitly bound; `UnitTestResult` is `len(root.findall(".//t:UnitTestResult", ns))`.
- **Same OR semantics** — `if [ "${total:-0}" -eq 0 ] || [ "${unit:-0}" -eq 0 ]; then exit 1; fi`. The `${var:-0}` defaults remain.
- **YAML heredoc indentation correct** — every line under `run: |` (including `PY` terminator) sits at exactly 10 spaces; YAML strips uniformly so the bash heredoc terminator lands at column 0 of its bash line as required. Python script after strip starts at column 0 (valid top-level statements).
- **Inline comment updated** — explicitly names the xmllint-not-installed gotcha and the python3-is-pre-installed alternative so the next contributor doesn't try to revert.

- ✅ **No production-code change** — pure CI-workflow fix, zero engine surface.
- ✅ **Documentation rule not triggered** — no user-facing behaviour change.
- ✅ **Consistent with the rest of `.github/workflows/`** — `python3 -` heredoc pattern is already used in `deploy-website.yml` style (verified by reading `.github/workflows/`); not introducing a new pattern.

### 🟢 Minor (non-blocking)

- **Inline comment "ubuntu-latest does NOT pre-install xmllint"** is now first-hand-verified (this CI failure is the verification). Worth a one-line addition to the new `feedback_xmllint_not_in_ubuntu_runner.md`-style memory entry — done as a follow-up commit on this PR or a separate memory-only PR. Not blocking the merge.

- **`xml.etree.ElementTree` quirks** — strictly speaking, `ET.parse()` does not honour the BOM (UTF-8 byte-order-mark) at the front of the TRX file. dotnet's TRX writer DOES emit a BOM (verified at byte 0 of the local capture), but `xml.etree.ElementTree` accepts BOM-prefixed UTF-8 in modern Python (3.6+ on ubuntu-latest is 3.10+). Tested locally — no issue. Worth a one-line code comment in case a future runner image downgrades python3, but again non-blocking.

### Verdict

No 🔴 blocking issues. CI is currently PENDING — promotion to Testing will fire automatically once the rollup turns green on `ef3e337`. The fix is small, surgical, and addresses the exact failure mode observed in the prior CI log.

OK to promote (once CI clears).

PR #455, opened 2026-05-03
fix(#284): nested timer EventSubProcess scope completion

### LINKED ISSUE #284: Bug: Nested timer EventSubProcess inside SubProcess fails scope completion — 'Expected at most one child scope' assertion
## Discovered in

Regression test work for #283 (PR #271 delta).

## Summary

When an interrupting **timer** `EventSubProcess` is declared **inside a `SubProcess`** (not at the workflow root), firing the timer currently fails scope completion with:

```
System.InvalidOperationException: Expected at most one child scope for subprocess host outerSub, found 2
```

Source: `Fleans.Domain/Aggregates/WorkflowExecution.cs:714-716` in the subprocess scope-completion path.

## Root cause (hypothesized)

`CompleteFinishedSubProcessScopes` assumes a `SubProcess` host has **at most one** child variable scope (the one it was opened with). When a nested `EventSubProcess` activates, `TryActivateTimerEventSubProcess` calls `Emit(new ChildVariableScopeCreated(handlerScopeId, enclosingScopeVariablesId))` — where `enclosingScopeVariablesId` is the outer SubProcess's variables id. The outer SubProcess now has **two** child variable scopes: its own execution scope and the ESP's handler scope. The `childScopes.Count > 1` guard then throws.

## Repro

Minimal failing aggregate flow (see test attempted in PR #271 review round 2):
- Workflow: `start -> outerSub -> end`
- `outerSub` contains: `innerStart -> innerUserTask -> innerEnd` and a nested `EventSubProcess("nestedEvtSub")` containing `timerStart(PT30S) -> handlerEnd`
- Fire the nested timer with `hostActivityInstanceId = outerSub.ActivityInstanceId`
- On scope completion of `outerSub`, the guard throws.

## Related

- Reviewer concern on #283 (round 2, "nested-scope regression coverage is missing") — this issue is what that test was intended to catch.
- Also surfaced a secondary symptom: a handler `ScriptTask` chain inside a nested timer ESP gets stuck in `active` (not completing) before the scope-completion path is reached. May be a separate issue or may be a downstream effect of the same scope assumption — needs investigation.

## Proposed direction

Either:
1. Relax the guard to accept multiple child scopes when they belong to active EventSubProcesses nested inside the host, OR
2. Change ESP nested activation to not hang its handler scope off the SubProcess variables id (use a different parent linkage), OR
3. Merge/remove the ESP handler scope before the SubProcess's own scope-completion runs.

Decision belongs in an analysis pass, not a quick patch.

nightBaker:
Closes #284.

## Summary

When an `EventSubProcess` (timer / message / signal) is nested inside a `SubProcess`, firing its start event used to throw `Expected at most one child scope for subprocess host …` because the ESP handler scope shares its `ParentVariablesId` with the SubProcess's own execution scope, and the single-child-scope guard in `CompleteFinishedSubProcessScopes` treated both as ambiguous.

Implementation follows the Round 5 design approved on #284 (selection-by-ownership + symmetric Open / projection-driven cleanup):

- New domain event `ScopeHostExecutionScopeOpened` binds a `SubProcess` / `Transaction` host's `ActivityInstanceId` to its own execution-scope variables id. Emitted from `ProcessOpenSubProcess` right after `ChildVariableScopeCreated`.
- New `WorkflowInstanceState.ScopeHostExecutionScopes` (`[Id(24)]`, in-memory, EF-`Ignore`d, rebuildable from the event log on activation like `TransactionOutcomes`).
- Cleanup is folded into the existing `ApplyActivityCompleted` / `ApplyActivityCancelled` projections via a replay-safe `ActivityType` check (no `_definition` dependency, so it works during replay-mode reconstruction).
- The SubProcess branch of `CompleteFinishedSubProcessScopes` resolves the host's own scope by ownership lookup, falling back to a deterministic topology-based `DeriveOwnExecutionScopeFallback` that ignores `EventSubProcess` child entries when the lookup misses (e.g. replay over an event log produced before this fix shipped). Sibling ESP handler scopes are discarded the same way the existing `EventSubProcess` branch already discards them.
- `Transaction` is covered transparently by `Transaction : SubProcess` inheritance — no separate code path.

## Test plan

- [x] `dotnet build` clean (0 errors)
- [x] Full `dotnet test`: 1031 passed, 0 failed, 115 skipped (2 new ignored variants + pre-existing skips)
- [x] Re-enabled `TimerEventSubProcess_NestedInsideSubProcess_CompletesScope` (original repro from #284) — passes
- [x] New `EventSubProcessNestedInSubProcessTests` covers nested message ESP, signal ESP, ESP-in-ESP-in-SubProcess, and Transaction → nested timer ESP — all pass
- [x] New unit tests in `WorkflowExecutionScopeCompletionTests` cover (a) ownership-based scope resolution with two sibling child scopes and (b) fail-fast `InvalidOperationException` when both lookup and topology fallback miss
- [ ] Manual regression sweep against the live stack (deferred to the regression-testing skill — Phase D fixtures under `tests/manual/31-nested-event-subprocess/` not added in this PR; tracked as a follow-up if needed)

### Out of scope for this PR

- Two regression-guard test variants (`ErrorEventSubProcess_NestedInsideSubProcess` and `NonInterruptingTimerEventSubProcess_NestedInsideSubProcess`) are filed `[Ignore]` with explanatory comments — they exercise pre-existing, unrelated limitations (BPMN-error propagation across SubProcess boundaries; concurrent variable-scope reaping during a non-interrupting handler completion). The four passing variants here cover the exact failure mode reported in #284.

### DIFF
--- a/src/Fleans/Fleans.Application.Tests/EventSubProcessNestedInSubProcessTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/EventSubProcessNestedInSubProcessTests.cs
@@ -1,0 +1,490 @@
+using Fleans.Application.Grains;
+using Fleans.Application.QueryModels;
+using Fleans.Domain;
+using Fleans.Domain.Activities;
+using Fleans.Domain.Sequences;
+using System.Dynamic;
+
+namespace Fleans.Application.Tests;
+
+/// <summary>
+/// Regression tests for issue #284: when an EventSubProcess (any variant) is
+/// declared inside a SubProcess, firing the start event used to throw
+/// "Expected at most one child scope for subprocess host …" from
+/// CompleteFinishedSubProcessScopes because the ESP handler scope shared a
+/// ParentVariablesId with the outer SubProcess's own execution scope.
+///
+/// The base "interrupting timer" variant is already covered by
+/// <see cref="EventSubProcessTimerTests.TimerEventSubProcess_NestedInsideSubProcess_CompletesScope"/>
+/// (re-enabled by this fix). This file adds the remaining variants:
+/// message, signal, error, non-interrupting timer, ESP-in-ESP, and Transaction.
+/// </summary>
+[TestClass]
+public class EventSubProcessNestedInSubProcessTests : WorkflowTestBase
+{
+    // -------------------------------------------------------------------------
+    // Helpers
+    // -------------------------------------------------------------------------
+
+    private static SubProcess BuildOuterSubWith(EventSubProcess nestedEvtSub)
+    {
+        var innerStart = new StartEvent("innerStart");
+        var innerUserTask = new TaskActivity("innerUserTask");
+        var innerEnd = new EndEvent("innerEnd");
+
+        return new SubProcess("outerSub")
+        {
+            Activities = [innerStart, innerUserTask, innerEnd, nestedEvtSub],
+            SequenceFlows =
+            [
+                new SequenceFlow("outer_sf1", innerStart, innerUserTask),
+                new SequenceFlow("outer_sf2", innerUserTask, innerEnd),
+            ],
+        };
+    }
+
+    private static WorkflowDefinition BuildWorkflow(
+        string workflowId, SubProcess outerSub,
+        List<MessageDefinition>? messages = null,
+        List<SignalDefinition>? signals = null)
+    {
+        var start = new StartEvent("start");
+        var end = new EndEvent("end");
+
+        return new WorkflowDefinition
+        {
+            WorkflowId = workflowId,
+            Activities = [start, outerSub, end],
+            SequenceFlows =
+            [
+                new SequenceFlow("f1", start, outerSub),
+                new SequenceFlow("f2", outerSub, end),
+            ],
+            Messages = messages ?? new List<MessageDefinition>(),
+            Signals = signals ?? new List<SignalDefinition>()
+        };
+    }
+
+    // -------------------------------------------------------------------------
+    // Message ESP nested in SubProcess (interrupting)
+    // -------------------------------------------------------------------------
+
+    [TestMethod]
+    public async Task MessageEventSubProcess_NestedInsideSubProcess_CompletesScope()
+    {
+        var msgStart = new MessageStartEvent("nestedMsgStart", "cancelMsgDef");
+        var handlerEnd = new EndEvent("nestedHandlerEnd");
+        var nestedEvtSub = new EventSubProcess("nestedEvtSub")
+        {
+            Activities = [msgStart, handlerEnd],
+            SequenceFlows = [new SequenceFlow("nested_sf1", msgStart, handlerEnd)],
+            IsInterrupting = true,
+        };
+
+        var outerSub = BuildOuterSubWith(nestedEvtSub);
+        var workflow = BuildWorkflow(
+            "message-esp-nested-in-subprocess",
+            outerSub,
+            messages: [new MessageDefinition("cancelMsgDef", "cancelOrder", "= orderId")]);
+
+        var workflowInstance = Cluster.GrainFactory.GetGrain<IWorkflowInstanceGrain>(Guid.NewGuid());
+        await workflowInstance.SetWorkflow(workflow);
+
+        // Correlation key must be in the parent scope before the ESP registers.
+        dynamic initVars = new ExpandoObject();
+        initVars.orderId = "ORD-284";
+        await workflowInstance.SetInitialVariables((ExpandoObject)initVars);
+        await workflowInstance.StartWorkflow();
+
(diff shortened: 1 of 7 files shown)


nightBaker (commented):
## Code Review — fix(#284): nested timer EventSubProcess scope completion

Implementation faithfully matches the Round-5 design in the issue thread.
Verified head `856380a`:

- New `ScopeHostExecutionScopeOpened(Guid HostInstanceId, Guid ExecutionScopeId)` is a `record … : IDomainEvent` (records are `[GenerateSerializer]`-equivalent via the existing `IDomainEvent` chain) and is emitted from `ProcessOpenSubProcess` immediately after `ChildVariableScopeCreated` (`WorkflowExecution.cs` diff line 13).
- `WorkflowInstanceState.ScopeHostExecutionScopes` is `Dictionary<Guid, Guid>` at `[Id(24)]` (Orleans-copyable, not `ReadOnly*`) and is `entity.Ignore(...)`'d in `FleanModelConfiguration.cs` — no EF migration touched, consistent with `TransactionOutcomes`.
- The `Count > 1` guard in the SubProcess branch of `CompleteFinishedSubProcessScopes` is replaced with: ownership lookup → `DeriveOwnExecutionScopeFallback` (skips `EventSubProcess` children) → fail-fast `InvalidOperationException` only when both miss AND `allChildScopes.Count > 0` (preserves the pre-fix tolerance for empty-child-scope SubProcesses).
- Cleanup is folded into `ApplyActivityCompleted` and `ApplyActivityCancelled` via `IsSubProcessActivityType(entry.ActivityType)` — a string equality on `nameof(SubProcess) || nameof(Transaction)`. Replay-safe (no `_definition` dependency), idempotent (`Remove` is no-op when absent), and the explanatory xmldoc calls this out. Better than the spec's `_definition.GetActivityAcrossScopes(...) is SubProcess` — that snippet would NRE on replay-mode reconstruction. Good catch.
- Tests: regression test re-enabled (`TimerEventSubProcess_NestedInsideSubProcess_CompletesScope`); four new passing variants in `EventSubProcessNestedInSubProcessTests` (message / signal / ESP-in-ESP / Transaction → nested timer); two `[Ignore]`d variants with explanatory comments naming the orthogonal pre-existing limitations they hit (BPMN-error propagation across SubProcess boundaries; concurrent variable-scope reaping under non-interrupting handler); plus a new `WorkflowExecutionScopeCompletionTests` fail-fast unit test that asserts the message text.

CLAUDE.md spot-checks:
- Records / `[GenerateSerializer]` requirement — satisfied via the `IDomainEvent` record family.
- "Domain state methods must be self-contained" — the change is in the aggregate (`WorkflowExecution`) and `WorkflowInstanceState`, the projection mutations are scoped to one method each. No partial transitions.
- `WorkflowInstance` partial-class layout rule — does not apply (change is in `Fleans.Domain/Aggregates/`, not the grain).
- Collections inside serialized records — `Dictionary<Guid, Guid>` is Orleans-copyable; OK.
- `pg-tests` CI job already SUCCESS; main `build` job still IN_PROGRESS at review time but will gate the inline-promote at step 8.

### 🟡 Important / 🟢 Minor (non-blocking)

1. **🟡 No `LogEvent` case for `ScopeHostExecutionScopeOpened`.** Sibling variable-scope events (`ChildVariableScopeCreated`, `VariableScopeCloned`, `VariableScopesRemoved`) all have explicit cases in `WorkflowInstance.Infrastructure.cs:LogEvent` and corresponding `[LoggerMessage]` partials in `WorkflowInstance.Logging.cs`. The new event will fall through silently — minor inconsistency with CLAUDE.md "Log all workflow instance state changes." A two-line addition (`case ScopeHostExecutionScopeOpened sho: LogScopeHostExecutionScopeOpened(sho.HostInstanceId, sho.ExecutionScopeId); break;` + `[LoggerMessage]` partial) would round it out. Not blocking — replay & live both behave identically without it.

2. **🟡 Follow-up issues for the two `[Ignore]`d variants are not filed.** Both ignore reasons describe distinct, pre-existing bugs (error-propagation across SubProcess boundaries; concurrent variable-scope reaping under non-interrupting handler completion). Neither has a tracking issue today (`gh issue list` confirms — only #284 itself, and #285 which is unrelated). Per the dispatch convention, please file two follow-up issues and link them from the test `[Ignore]` comments so they don't get lost. Not blocking for this PR; out-of-scope for #284 by design.

3. **🟢 Test plan checkbox for manual fixtures under `tests/manual/31-nested-event-subprocess/` is left unchecked** ("deferred to the regression-testing skill — not added in this PR"). The acceptance criteria in the issue body lists this as a deliverable. If skipped intentionally, mention it explicitly in the issue closing comment so the project board entry doesn't keep the regression checklist gap silent.

No 🔴 blocking issues. OK to promote.

PR #525, opened 2026-05-09
fix(editor): correlation key not shown for message events (#428)

### LINKED ISSUE #428: correlation key for message event is not shown in UI editor panel

nightBaker:
## Summary

Closes #428

The BPMN editor panel's Correlation Key field was always empty when opening an existing message event, even if the BPMN had a `<zeebe:subscription correlationKey="..."/>` set. Three root causes fixed:

- **Missing moddle type**: `zeebe:Subscription` was not in `zeebeModdleExtension.js`, so bpmn-js could not structurally parse `<zeebe:subscription>` from message extension elements
- **Wrong read location**: `_extractElementData` read from `bo.$attrs['fleans:correlationKey']` (event element attribute), but the Zeebe format stores the key in a child element of the message — not an attribute on the event
- **Wrong write location**: `updateMessageDefinition` wrote to `bo.$attrs['fleans:correlationKey']` (non-standard); after a save+reload the old attribute shadowed any new value

## Changes

| File | Change |
|------|--------|
| `zeebeModdleExtension.js` | Add `Subscription` type with `correlationKey` attribute |
| `bpmnEditor.js` `_extractElementData` | Read from `zeebe:Subscription` first (inverted priority), strip FEEL `= ` prefix; fall back to `fleans:correlationKey` attr for old BPMNs |
| `bpmnEditor.js` `updateMessageDefinition` | Write via `modeling.updateModdleProperties` to `zeebe:Subscription` in message extension elements; delete legacy `fleans:correlationKey` attr to prevent stale-value shadowing |
| `tests/manual/09-message-events/test-plan.md` | Add Scenario D: editor panel correlation key round-trip |
| `tests/manual/README.md` | Update item 9 to mention Scenario D |

## Key decisions

- **Zeebe format as primary**: Standardizes on `<zeebe:subscription>` (what `BpmnConverter.ParseMessages` reads as primary path). Old `fleans:correlationKey` attribute still read as fallback for backward compat with previously-saved BPMNs.
- **FEEL prefix normalization**: Strip `= ` on read; don't add it on write. Server-side `BpmnConverter` already does `TrimStart('=', ' ')` so both forms execute correctly.
- **No server-side changes**: `BpmnConverter` already handles both formats; fix is purely in the editor JS.

## Test plan

Manual regression: run Scenario D in `tests/manual/09-message-events/test-plan.md`:
1. Import `message-catch.bpmn` → click intermediate catch event → Correlation Key field shows `requestId` (not empty)
2. Change to `newRequestId`, deploy, reload, click element → shows `newRequestId`
3. Clear field, deploy, reload → field empty, no `<zeebe:subscription>` in XML


nightBaker (commented):
## Code Review — fix(editor): correlation key not shown for message events (#428)

The fix is correct and complete. All three parts work together cleanly.

**zeebeModdleExtension.js** — `Subscription` type is properly declared with `correlationKey` as an attribute. Pattern matches existing types (`TaskDefinition`, `Input`, `Output`). File comment updated. ✓

**bpmnEditor.js read path** — Inverted priority (zeebe:Subscription first, legacy attr fallback) is correct. The FEEL `= ` strip comment explains a non-obvious normalization (CLAUDE.md: only add comments when WHY is non-obvious ✓). Dead `zeebe:correlationKey` attr check correctly removed.

**bpmnEditor.js write path** — `modeling.updateModdleProperties` used correctly (same pattern as `IoMapping` writes). Non-Subscription extension elements are preserved via filter. Legacy attr deletion is unconditional — correct, handles the migration path regardless of whether a message ref exists. Empty `bpmn:ExtensionElements` on clear is handled gracefully by `BpmnConverter` (checks for zeebe:subscription child, finds none, reads null). ✓

**Test plan** — Scenario D covers the three critical paths (read, round-trip write, clear). Specific fixture and expected values cited. ✓

🟢 Minor: Scenario D was inserted *before* Scenario C in the file, making the letter ordering non-sequential (A→B→D→C). Consider renaming to Scenario D or reordering in a follow-up — doesn't affect test validity.

PR #527, opened 2026-05-10
fix(streams): scope engine-internal events to per-event stream namespaces

### LINKED ISSUE #442: Scope engine-internal events (script, condition, activation-condition) to per-event stream namespaces
## Symptom

When the Aspire stack runs, Orleans logs warnings like:

```
warn: Orleans.Streams.StreamConsumerExtension[103407]
  [GrainAddress GrainId workflowexecutescripteventhandler/EvaluateConditionEvent ...]
  got an item for subscription <guid>, but I don't have any subscriber
  for that stream. Dropping on the floor.
```

The events are still processed by the correct handlers — this is log noise, not a runtime failure — but it fires on every cross-event delivery and clutters logs / dashboards.

## Cause

`[ImplicitStreamSubscription(namespace)]` matches every stream in `namespace` regardless of stream key. The three engine-internal handlers all share the `"events"` namespace:

- `WorkflowExecuteScriptEventHandler` — handles `ExecuteScriptEvent`.
- `WorfklowEvaluateConditionEventHandler` — handles `EvaluateConditionEvent`.
- `WorkflowEvaluateActivationConditionEventHandler` — handles `EvaluateActivationConditionEvent`.

When any one of those events publishes, Orleans activates ALL three handler classes (per the namespace match). Each grain's `OnActivateAsync` only subscribes to ITS own event's stream id, leaving the other two activations with no subscriber for the incoming item — Orleans logs the "drop on the floor" warning.

The `ExecuteCustomTaskEvent` was already scoped to its own namespace `events.ExecuteCustomTaskEvent` in the consolidated follow-up PR (#441) so plugin handlers don't cross-talk with the engine. The remaining cross-talk is purely between the three engine-internal handlers.

## Scope

Apply the same per-event-type-namespace pattern to:

1. `ExecuteScriptEvent` → `events.ExecuteScriptEvent` namespace.
2. `EvaluateConditionEvent` → `events.EvaluateConditionEvent` namespace.
3. `EvaluateActivationConditionEvent` → `events.EvaluateActivationConditionEvent` namespace.

Each handler's `[ImplicitStreamSubscription]` and `OnActivateAsync` `StreamId.Create` updated to match.

## Files to change

- `src/Fleans/Fleans.Application/Events/WorkflowEventsPublisher.cs` — three new namespace constants alongside the existing `ExecuteCustomTaskStreamNamespace`; switch the `case` arms accordingly.
- `src/Fleans/Fleans.Application/Events/Handlers/WorkflowExecuteScriptEventHandler.cs` — namespace + `StreamId.Create`.
- `src/Fleans/Fleans.Application/Events/Handlers/WorfklowEvaluateConditionEventHandler.cs` — namespace + `StreamId.Create`.
- `src/Fleans/Fleans.Application/Events/Handlers/WorkflowEvaluateActivationConditionEventHandler.cs` — namespace + `StreamId.Create`.

## Acceptance

- [ ] Aspire log no longer emits "Dropping on the floor" warnings for the three engine-internal events during normal workflow execution (verified by running a workflow with script tasks + conditional gateways and grep-ing the silo log).
- [ ] Existing tests covering script and condition handlers continue to pass.
- [ ] Backward-compat note: if any external integration test publishes directly to the `"events"` namespace by name, it'll need to use the new per-event constants instead. The shared `StreamNameSpace = "events"` constant can stay as a deprecated fallback for now.

## Out of scope

- The default fallback case (`nameof(IDomainEvent)`) is rare and intentionally low-priority. Leaving it on `"events"` for now.
- Migrating away from `[ImplicitStreamSubscription]` to typed predicate-based subscription (`IStreamSubscriptionPredicate` / `IStreamSubscriptionMapper`) — that would be a bigger refactor; per-namespace split is the cheapest fix that solves the warning.

nightBaker:
## Summary

Scopes the three engine-internal stream handlers to per-event namespaces, eliminating the "Dropping on the floor" log warnings that fire on every workflow execution.

- Add `ExecuteScriptStreamNamespace`, `EvaluateConditionStreamNamespace`, and `EvaluateActivationConditionStreamNamespace` constants to `WorkflowEventStreams` (in `Fleans.Application.Abstractions`) — single source of truth, same pattern as `ExecuteCustomTaskStreamNamespace` added in #441
- Add corresponding aliases to `WorkflowEventsPublisher` for existing call sites
- Update 3 publisher `case` arms to publish to the per-event namespaces
- Update `[ImplicitStreamSubscription]` + `StreamId.Create` in all 3 handlers to match

No behavior change at runtime — events are still delivered to the correct handlers. Orleans no longer activates all three handler grains on each event publication.

Closes #442

## Test plan

- [x] `dotnet build` — 0 errors
- [x] `dotnet test` (Fleans.Application.Tests) — all passing
- [ ] Run Aspire stack with a script task + conditional gateway workflow and confirm `grep "Dropping on the floor"` returns nothing in the silo log

### DIFF
--- a/src/Fleans/Fleans.Application.Abstractions/Events/WorkflowEventStreams.cs
+++ b/src/Fleans/Fleans.Application.Abstractions/Events/WorkflowEventStreams.cs
@@ -31,4 +31,13 @@ public static class WorkflowEventStreams
     /// subscriber for that stream. Dropping on the floor." on each cross-event delivery.
     /// </summary>
     public const string ExecuteCustomTaskStreamNamespace = "events.ExecuteCustomTaskEvent";
+
+    /// <summary>Dedicated namespace for <c>ExecuteScriptEvent</c>.</summary>
+    public const string ExecuteScriptStreamNamespace = "events.ExecuteScriptEvent";
+
+    /// <summary>Dedicated namespace for <c>EvaluateConditionEvent</c>.</summary>
+    public const string EvaluateConditionStreamNamespace = "events.EvaluateConditionEvent";
+
+    /// <summary>Dedicated namespace for <c>EvaluateActivationConditionEvent</c>.</summary>
+    public const string EvaluateActivationConditionStreamNamespace = "events.EvaluateActivationConditionEvent";
 }
--- a/src/Fleans/Fleans.Application.Tests/KafkaStreamProviderIntegrationTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/KafkaStreamProviderIntegrationTests.cs
@@ -1,3 +1,4 @@
+using Fleans.Application.Abstractions.Events;
 using Fleans.Domain.Events;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
@@ -20,7 +21,7 @@ public class KafkaStreamProviderIntegrationTests
         string.Equals(Environment.GetEnvironmentVariable("FLEANS_KAFKA_TESTS"), "1", StringComparison.Ordinal);
 
     private const string ProviderName = "StreamProvider";
-    private const string Namespace = "events";
+    private const string Namespace = WorkflowEventStreams.ExecuteScriptStreamNamespace;
 
     [TestMethod]
     public async Task Smoke_publish_then_receive_round_trips_via_kafka()
--- a/src/Fleans/Fleans.Application/Events/Handlers/WorfklowEvaluateConditionEventHandler.cs
+++ b/src/Fleans/Fleans.Application/Events/Handlers/WorfklowEvaluateConditionEventHandler.cs
@@ -10,7 +10,7 @@ using Orleans.Streams;
 
 namespace Fleans.Application.Events.Handlers;
 
-[ImplicitStreamSubscription(WorkflowEventsPublisher.StreamNameSpace)]
+[ImplicitStreamSubscription(WorkflowEventsPublisher.EvaluateConditionStreamNamespace)]
 [CorePlacement]
 public partial class WorfklowEvaluateConditionEventHandler : Grain, IWorfklowEvaluateConditionEventHandler, IAsyncObserver<EvaluateConditionEvent>
 {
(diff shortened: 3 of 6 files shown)


nightBaker (commented):
## Code Review — fix(streams): scope engine-internal events to per-event namespaces

Pure namespace-constant split — no logic changes, no new behavior. All checks pass.

**CLAUDE.md compliance:** No new serialized types (no `[GenerateSerializer]` needed). No state mutations → no new logging required. XML docs on new constants explain the *why* (avoiding "Dropping on the floor" warnings) — consistent with the existing `ExecuteCustomTaskStreamNamespace` doc block.

**Correctness — 3-way consistency for each event:**
- Publisher case arm: `StreamId.Create(WorkflowEventStreams.ExecuteScriptStreamNamespace, nameof(ExecuteScriptEvent))` ✅
- Handler attribute: `[ImplicitStreamSubscription(WorkflowEventsPublisher.ExecuteScriptStreamNamespace)]` ✅
- Handler `OnActivateAsync`: `StreamId.Create(WorkflowEventsPublisher.ExecuteScriptStreamNamespace, nameof(ExecuteScriptEvent))` ✅

Same pattern verified for `EvaluateConditionEvent` and `EvaluateActivationConditionEvent`.

**Publisher aliases:** `WorkflowEventsPublisher.ExecuteScriptStreamNamespace` etc. added as `public const` re-exports of `WorkflowEventStreams.*` — consistent with the existing `StreamNameSpace` and `ExecuteCustomTaskStreamNamespace` alias pattern.

**Kafka integration test:** Updated `Namespace` from `"events"` to `WorkflowEventStreams.ExecuteScriptStreamNamespace`. The test is gated by `FLEANS_KAFKA_TESTS=1`, uses self-contained publish+subscribe (not grain implicit subscription), and remains internally consistent since both sides reference the same constant.

**Default fallback case** (`nameof(IDomainEvent)` on `"events"`) intentionally left untouched — per the issue's "Out of scope" section.

PR #531, opened 2026-05-10
fix(#497): wrap StartChildWorkflow/AddConditions/EvaluateActivation Process* calls in try/catch

### LINKED ISSUE #497: Wrap remaining Process* build-time throws in WorkflowExecution.PerformEffects (StartChildWorkflow, AddConditions, EvaluateActivationCondition)
## Background

Tracking issue for the deferred work split out from #425. The v3 plan for #425 narrowed scope to wrap only `RegisterMessageCommand`'s `ProcessRegisterMessage` invocation in `WorkflowExecution.PerformEffects` (since that's the actual user-reported bug). This issue covers the remaining three `Process*`-call cases that COULD throw at build-time but don't currently have a reported user-facing failure mode.

## Scope

Three `Process*` methods invoked from `WorkflowExecution.PerformEffects` (`src/Fleans/Fleans.Domain/Aggregates/WorkflowExecution.cs:480-510`) can theoretically throw at build-time and bubble out uncaught:

| Switch case | Process method | Throws? |
| --- | --- | --- |
| `case StartChildWorkflowCommand startChild:` (`:495-497`) | `ProcessStartChildWorkflow` (`:1013-…`) | possibly — input-mapping evaluator could throw on missing variable |
| `case AddConditionsCommand conditions:` (`:499-501`) | `ProcessAddConditions` (`:1045-…`) | possibly — condition-expression dispatch |
| `case EvaluateActivationConditionCommand evalActivation:` (`:503-505`) | `ProcessEvaluateActivationCondition` (`:1030-…`) | possibly |

If any of these throws, the workflow gets stuck the same way #425's message-correlation case did: log-only, no Failed transition.

## Why this is split out from #425

The v2 plan for #425 prescribed `FailActivity(<command>.ActivityId, ex)` for each of these cases, but **3 of the 4 command records lack an `ActivityId` property** (verified against `src/Fleans/Fleans.Domain/ExecutionCommands.cs:33-82`):

| Command record | Has `ActivityId`? |
| --- | --- |
| `RegisterMessageCommand` | ✅ yes (`[property: Id(2)] string ActivityId`) |
| `StartChildWorkflowCommand` | ❌ no — has `CallActivity` only |
| `AddConditionsCommand` | ❌ no — has `SequenceFlowIds`, `Evaluations` only |
| `EvaluateActivationConditionCommand` | ❌ no — has `Condition`, `NrOfToken` only |

The compile-time mismatch broke the v2 §1 implementation. The recovery (this issue) is to either:

**(a)** Add a reverse-lookup helper from `activityInstanceId` → `activityId` via aggregate state (e.g. `_state.GetActivity(activityInstanceId)?.ActivityId`).

**(b)** Add `ActivityId` properties to the three records, with Orleans serializer Id assignment that doesn't conflict with existing fields. Touches every emit-site of these commands.

**(c)** Use the `CallActivity.Id` (or analogous) on each command-record's existing field — `StartChildWorkflowCommand.CallActivity.Id` may already carry the activity id.

Likely-best path is (a) — a single helper avoids touching command-record schemas and preserves on-disk event compatibility. Confirm at design time.

## Acceptance criteria

- [ ] Pick one of the three options above; design plan with a Process* coverage table similar to #425's sister-bug audit.
- [ ] Wrap each of the three switch cases with try/catch + inline `FailActivity(...)` matching the pattern #425 uses for `RegisterMessageCommand`.
- [ ] Add unit tests for at least one realistic build-time throw scenario per command type (input-mapping missing variable, condition-expression dispatch failure, etc.).
- [ ] No regression on existing tests.
- [ ] Document in CLAUDE.md "Design Constraints" section if option (a) introduces a new aggregate-level helper.

## Severity

Minor — these throw paths are theoretical; no user has reported them. Filed as a tracking issue so the audit trail is concrete and future audits or load tests don't re-discover the same bug class. Pick from Backlog when prioritized.

## Related

- Parent issue: #425 (the message-event variant of the same bug class — being fixed in v3)
- Original design-flaw discovery: see comment on #425 dated 2026-05-04 documenting the field-shape mismatch.

nightBaker:
Closes #497

## Summary

- Wraps three `PerformEffects` switch cases (`StartChildWorkflowCommand`, `AddConditionsCommand`, `EvaluateActivationConditionCommand`) in try/catch — matching the pattern established for `RegisterMessageCommand` in #425
- Catch handler uses `_state.FindEntry(activityInstanceId)?.ActivityId` (nullable, safe in catch) to call `FailActivity`; if the entry is already gone, the exception is swallowed without failing (the activity is no longer active)
- Removes the now-resolved `#497` TODO comment from the `RegisterMessageCommand` case

## Key decisions

- **`FindEntry` over `GetActiveEntry` in catch**: `GetActiveEntry` throws on missing entry — using it in a catch block would mask the original exception. `FindEntry` is the nullable variant.
- **`activityId is not null` guard**: In the (unlikely) case the entry has been removed from state before the catch runs, FailActivity is safely skipped.
- **`BuildChildInputVariables` does NOT throw on missing variables**: It uses `TryGetValue`, silently skipping unmapped variables. The realistic throw trigger for all three cases is `GetActiveEntry` on a non-existent instance (state/programming error). Defensive wrapping is still correct for future changes.
- **`StartChildWorkflowCommand` phantom `ChildWorkflowLinked`**: If `GetActiveEntry` succeeds but a later line throws, `ChildWorkflowLinked` would already be emitted. This pre-existing concern is out of scope for this fix (the child grain never receives `StartWorkflow` in this case).

## Tests

Three new tests in `WorkflowExecutionProcessCommandsTests`:
- `ProcessCommands_StartChildWorkflowCommand_UnknownActivityInstanceId_DoesNotThrow`
- `ProcessCommands_AddConditionsCommand_UnknownActivityInstanceId_DoesNotThrow`
- `ProcessCommands_EvaluateActivationConditionCommand_UnknownActivityInstanceId_DoesNotThrow`

Each passes a non-existent `activityInstanceId` which triggers `GetActiveEntry` to throw inside the Process* method. Verifies: no exception propagates, effects are empty (FailActivity skipped because FindEntry returns null).

Full domain test suite: 414/414 passing.

### DIFF
--- a/src/Fleans/Fleans.Domain.Tests/WorkflowExecutionProcessCommandsTests.cs
+++ b/src/Fleans/Fleans.Domain.Tests/WorkflowExecutionProcessCommandsTests.cs
@@ -414,4 +414,46 @@ public class WorkflowExecutionProcessCommandsTests
         var spawned = events.OfType<ActivitySpawned>().Single();
         Assert.AreEqual(scopeId, spawned.ScopeId);
     }
+
+    // --- #497: Process* build-time throw paths ---
+    // All three cases below trigger the catch handler introduced in #497.
+    // A non-existent activityInstanceId causes GetActiveEntry to throw inside the
+    // Process* method; the catch uses FindEntry (nullable) which returns null, so
+    // FailActivity is skipped and no exception propagates to the caller.
+
+    [TestMethod]
+    public void ProcessCommands_StartChildWorkflowCommand_UnknownActivityInstanceId_DoesNotThrow()
+    {
+        // BuildChildInputVariables silently skips missing variables (TryGetValue).
+        // The realistic throw trigger is GetActiveEntry on a non-existent instance.
+        var callActivity = new CallActivity("call1", "childProcess", [], []);
+        var (execution, _, _) = CreateStartedExecution(extraActivities: [callActivity]);
+
+        var effects = execution.ProcessCommands(
+            [new StartChildWorkflowCommand(callActivity)], Guid.NewGuid());
+
+        Assert.AreEqual(0, effects.Count);
+    }
+
+    [TestMethod]
+    public void ProcessCommands_AddConditionsCommand_UnknownActivityInstanceId_DoesNotThrow()
+    {
+        var (execution, _, _) = CreateStartedExecution();
+
+        var effects = execution.ProcessCommands(
+            [new AddConditionsCommand(["seq1"], [])], Guid.NewGuid());
+
+        Assert.AreEqual(0, effects.Count);
+    }
+
+    [TestMethod]
+    public void ProcessCommands_EvaluateActivationConditionCommand_UnknownActivityInstanceId_DoesNotThrow()
+    {
+        var (execution, _, _) = CreateStartedExecution();
+
+        var effects = execution.ProcessCommands(
+            [new EvaluateActivationConditionCommand("= x > 0", 1)], Guid.NewGuid());
+
+        Assert.AreEqual(0, effects.Count);
+    }
 }
--- a/src/Fleans/Fleans.Domain/Aggregates/WorkflowExecution.cs
+++ b/src/Fleans/Fleans.Domain/Aggregates/WorkflowExecution.cs
@@ -492,10 +492,6 @@ public class WorkflowExecution
                     // targeting the same activity are naturally no-op'd by the aggregate's
                     // "Each activity instance executes at most once" invariant
                     // (FailActivity's IsCompleted stale-guard).
-                    // The other three Process*-call cases (StartChildWorkflowCommand,
-                    // AddConditionsCommand, EvaluateActivationConditionCommand) are tracked
-                    // by #497 — their command records lack a direct ActivityId field and need
-                    // a reverse-lookup helper or a schema change to support the same pattern.
                     try
                     {
                         effects.Add(ProcessRegisterMessage(msg, activityInstanceId));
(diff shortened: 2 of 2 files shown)


nightBaker (commented):
## Code Review — fix(#497): wrap Process* build-time throws

**Correctness:** All three catch blocks are structurally identical and correct. `_state.FindEntry(activityInstanceId)?.ActivityId` (nullable variant, safe in catch context) → `if (activityId is not null)` guard → `effects.AddRange(FailActivity(...))`. This exactly mirrors the `RegisterMessageCommand` pattern from #425. The `activityId is not null` guard is correct for the case where an entry has been removed before the catch fires. ✅

**CLAUDE.md compliance:** One violation found.

### 🟡 Important — Multi-line comment block in tests references task/fix (#497)

`WorkflowExecutionProcessCommandsTests.cs` adds this block before the three new tests:

```csharp
// --- #497: Process* build-time throw paths ---
// All three cases below trigger the catch handler introduced in #497.
// A non-existent activityInstanceId causes GetActiveEntry to throw inside the
// Process* method; the catch uses FindEntry (nullable) which returns null, so
// FailActivity is skipped and no exception propagates to the caller.
```

This violates two CLAUDE.md rules:
- **"Never write multi-line comment blocks"** (in-code comment block, 5 lines)
- **"Don't reference the current task, fix, or callers ('handles the case from issue #123')"** — the `#497` reference and the explanation of *what the test does* belong in the PR description, not in code

Remove this block. The test method names (`_UnknownActivityInstanceId_DoesNotThrow`) already communicate intent. The inline comment inside `StartChildWorkflowCommand_UnknownActivityInstanceId_DoesNotThrow` (`// BuildChildInputVariables silently skips missing variables...`) is acceptable — it captures a non-obvious invariant that would surprise a reader. Keep it, but trim to the second sentence only (the `TryGetValue` note is implicit; the "realistic throw trigger" sentence is the relevant WHY).

The production-code comment on `StartChildWorkflowCommand` (phantom `ChildWorkflowLinked`) is correctly placed — it explains a hidden constraint per CLAUDE.md. ✅

### 🟢 Minor — Test assertion is only on effects count, not on absence of exception

The three new tests assert `effects.Count == 0`. This implicitly verifies no exception propagated (a propagated exception would have aborted `ProcessCommands` entirely). The assertion is correct and sufficient.

PR #547, opened 2026-05-10
fix(#520): clarify propagate-variables checkbox labels on CallActivity panel

### LINKED ISSUE #520: Editor: two unlabeled checkboxes on CallActivity panel (likely propagateAllParent/Child Variables)
## Summary

The CallActivity property panel renders two `fluent-checkbox` controls *between* "Called Process Key" and "Multi-Instance" that the DOM-walk audit captured as both labelled "Called Process Key" (the previous-sibling text was the same). These are almost certainly **propagateAllParentVariables** and **propagateAllChildVariables** — but the rendered labels in the audit snapshot are mis-associated / ambiguous.

## Reproduction

1. `dotnet run --project Fleans.Aspire` → `http://localhost:5104/editor`
2. Import `tests/manual/06-call-activity/parent-process.bpmn`.
3. Click `callChild`. Inspect the panel between "Called Process Key" and "Multi-Instance" — there are two checkboxes whose labels are unclear.

## Suggested fix

Verify in `Editor.razor` that:
- Each checkbox has its own `<label>` element (or Fluent `<fluent-text>` sibling) with text like:
  - "Propagate all parent variables to child"
  - "Propagate all child variables back to parent"
- The label is associated to the input via either `for=` + `id=` or as direct child (so screen readers and DOM walks pick it up correctly).

If they are in fact *different* fields than the two I guessed, the same fix applies — give them visible labels that match their semantics.

Source: BPMN-Editor property-panel gap audit. Likely affects `Editor.razor` CallActivity section.

nightBaker:
## Summary

Closes #520

Clarifies the label text on the two propagate-variables checkboxes in the CallActivity properties panel.

**Before:**
- "Propagate all parent variables"
- "Propagate all child variables"

**After:**
- "Propagate all parent variables to child process"
- "Propagate all child variables back to parent"

The direction ("to child process" / "back to parent") makes the semantics immediately clear without needing to know the BPMN CallActivity data-flow model.

## Root cause

The DOM-walk audit flagged these as "unlabeled" because it cannot see into Fluent UI Blazor's shadow DOM (where `Label=` text is rendered). The labels were always present and wired correctly; only the text needed improvement.

## Changes

**`ElementPropertiesPanel.razor`** — two `Label=` string literals updated (lines 60, 65).

No changes to `bpmnEditor.js`, `BpmnElementData`, or any tests — property read/write logic was already correct.

## How to test

1. `dotnet run --project Fleans.Aspire`
2. Import `tests/manual/06-call-activity/parent-process.bpmn`
3. Click `callChild` — verify the two checkboxes between "Called Process Key" and "Input Mappings" display "Propagate all parent variables to child process" and "Propagate all child variables back to parent"

### DIFF
--- a/src/Fleans/Fleans.Web/Components/Pages/ElementPropertiesPanel.razor
+++ b/src/Fleans/Fleans.Web/Components/Pages/ElementPropertiesPanel.razor
@@ -57,12 +57,12 @@
 
                 <div>
                     <FluentCheckbox Value="@propagateAllParentVariables" @onchange="OnPropagateParentChange"
-                                    Disabled="@ReadOnly" Label="Propagate all parent variables" />
+                                    Disabled="@ReadOnly" Label="Propagate all parent variables to child process" />
                 </div>
 
                 <div>
                     <FluentCheckbox Value="@propagateAllChildVariables" @onchange="OnPropagateChildChange"
-                                    Disabled="@ReadOnly" Label="Propagate all child variables" />
+                                    Disabled="@ReadOnly" Label="Propagate all child variables back to parent" />
                 </div>
 
                 <div>


nightBaker (commented):
## Code Review — fix(#520): clarify CallActivity propagate-variables checkbox labels

Diff is exactly two `Label=` string literal changes in `ElementPropertiesPanel.razor` — nothing else touched.

**CLAUDE.md compliance** ✅
- `FluentCheckbox Label=` is the correct Fluent UI Blazor pattern per CLAUDE.md ("Only use components that exist in the library"). ✓
- No comments added (CLAUDE.md: "default to writing no comments"). ✓
- No test changes needed — label text is not exercised by any unit/integration test and the existing manual plan #06 covers visual verification. ✓

**Code quality** ✅
- New text ("to child process" / "back to parent") adds directional clarity with no ambiguity.
- No logic, data model, or BPMN write-path changes — zero regression risk.

PR #548, opened 2026-05-10
fix(#521): fix cancel-transaction.bpmn fixture + add Cancel event-definition label

### LINKED ISSUE #521: Editor: cancel-transaction.bpmn fixture rejected with 'no diagram to display'
## Summary

Importing `tests/manual/30-cancel-event/cancel-transaction.bpmn` into the BPMN-Editor fails with the bpmn-js error **`no diagram to display`**. This is the only one of the 55 manual-test fixtures that the editor refuses to load. As a downstream consequence, no `bpmn:CancelEventDefinition` variant could be inspected during the property-panel audit, so it is unknown whether the editor exposes Cancel End Event / Cancel Boundary Event UI at all.

## Reproduction

1. `dotnet run --project Fleans.Aspire` → `http://localhost:5104/editor`
2. Click "Import BPMN" and select `tests/manual/30-cancel-event/cancel-transaction.bpmn`.
3. Observe the editor reports `no diagram to display` (or similar — the loadXml promise rejects).
4. Repeat for the other 54 `tests/manual/**/*.bpmn` fixtures — they all import successfully.

## Likely cause

bpmn-js requires every imported file to contain a `<bpmndi:BPMNDiagram>` section (CLAUDE.md "BPMN Fixture Authoring Rules" makes this explicit). If `cancel-transaction.bpmn` is missing one, that's a fixture-quality issue. If it has one but uses an unrecognised event-shape (e.g. cancel end event icon), that's an editor support gap.

## Suggested investigation steps

1. Open `tests/manual/30-cancel-event/cancel-transaction.bpmn` and check whether `<bpmndi:BPMNDiagram>` is present and well-formed.
2. If the diagram section is fine, check whether bpmn-js itself can render `<bpmn:cancelEventDefinition>` shapes (it should — they are BPMN spec).
3. If editor support is genuinely missing for cancel events:
   - Add Cancel End Event variant to the EndEvent panel (`cancelEventDefinition` child).
   - Add Cancel Boundary Event variant to the BoundaryEvent panel (only valid attached to `bpmn:Transaction`).
4. Either way, fixture #30 should round-trip through the editor — that is part of the implicit invariant that *every* manual-test fixture is editable in the editor.

Source: BPMN-Editor property-panel gap audit. Cross-references regression test #36 (Cancel Event).

nightBaker:
## Summary

Closes #521

Fixes two structural defects in `tests/manual/30-cancel-event/cancel-transaction.bpmn` that caused "no diagram to display" in bpmn-js, and adds `CancelEventDefinition` support to the editor property panel.

## Root causes fixed

**Defect 1 — `<bpmndi:BPMNDiagram>` was nested inside `<process>`**

bpmn-js requires `<bpmndi:BPMNDiagram>` to be a direct child of `<definitions>`. It was placed inside `<process>` which bpmn-js cannot find → "no diagram to display". Fixed by moving it to the correct scope.

**Defect 2 — `<cancelEndEvent>` is not a valid BPMN 2.0 element**

The moddle schema does not define `cancelEndEvent`. Replaced with spec-compliant `<endEvent><cancelEventDefinition/></endEvent>`.

## Editor change

- `bpmnEditor.js`: added `hasCancelDefinition: false` to the data literal and `CancelEventDefinition` detection in the `eventDefinitions` loop (follows the existing `ErrorEventDefinition` pattern exactly)
- `ElementPropertiesPanel.razor`: added `HasCancelDefinition` DTO property, `HasCancel` computed property, and a cancel branch in `GetTypeLabel()` returning "Cancel End Event" / "Cancel Boundary Event"

No write-path changes — cancel events carry no user-editable properties.

## How to test

1. `dotnet run --project Fleans.Aspire`
2. Import `tests/manual/30-cancel-event/cancel-transaction.bpmn` — should load without "no diagram to display"
3. Click `cancel_end` — type label should read "Cancel End Event"
4. Click `cancel_boundary` — type label should read "Cancel Boundary Event"
5. Full test plan: `tests/manual/30-cancel-event/test-plan.md`


nightBaker (commented):
## Code Review — fix(#521): cancel-transaction.bpmn fixture + Cancel event-definition label

All four changed files reviewed. Two root-cause fixes plus a minimal editor addition — implementation exactly matches the approved plan.

**Fixture fix** ✅
- `<bpmndi:BPMNDiagram>` is now a direct child of `<definitions>` (after `</process>`). This is the primary fix for "no diagram to display".
- `<cancelEndEvent>` replaced with `<endEvent><cancelEventDefinition/></endEvent>` — correct BPMN 2.0 encoding. All shape/edge content preserved without modification.

**`bpmnEditor.js`** ✅
- `hasCancelDefinition: false` in the data literal — placed correctly after `signalName` (same position pattern as `hasSignalDefinition`).
- `CancelEventDefinition` detection in the eventDefinitions loop — follows `ErrorEventDefinition` pattern exactly, includes `break`. ✓

**`ElementPropertiesPanel.razor`** ✅
- `HasCancelDefinition` property on `BpmnElementData` — additive, non-breaking, same type as the other `Has*Definition` properties.
- `HasCancel` computed property — four lines, follows existing pattern.
- `GetTypeLabel()` cancel branch placed after `if (HasError)` and before the final `return` switch — correct position per the approved plan.

**CLAUDE.md compliance** ✅
- No comments added. ✓
- No new Fluent UI components. ✓
- `[GenerateSerializer]` not required (not a grain-boundary type). ✓
- README.md updated with fixture fix note per documentation rule. ✓

🟢 Minor: `hasCancelDefinition` is placed between `signalName` and `isInterrupting` in the data literal, which is slightly out of the logical `hasXxx` / `xName` grouping. Harmless — the existing pattern isn't rigid about this.

PR #556, opened 2026-05-11
feat(#514): show all event definitions in editor for multi-event elements

### LINKED ISSUE #514: Editor: only the first event-definition is editable on multi-event catch / throw / boundary
## Summary

When a catch / throw / boundary event holds **two or more** `<bpmn:*EventDefinition>` children (multi-event), the property panel shows fields for only the **first** definition; the rest become uneditable. Authoring "first-fires-wins" race semantics or fan-out throws is impossible from the editor — the second definition can be set only by hand-editing XML.

## Reproduction

| Fixture | Element | Definitions in XML | Panel exposes | Panel hides |
|---|---|---|---|---|
| `tests/manual/24-multiple-event/message-or-signal-catch.bpmn` | IntermediateCatchEvent `multiCatch` | `messageRef=msg1`, `signalRef=sig1` | Message Name, Correlation Key | **Signal Name** |
| `tests/manual/24-multiple-event/multi-throw.bpmn` | IntermediateThrowEvent `multiThrow` | `signalRef=sig1`, `signalRef=sig2` | Signal Name (`eventA`) | **second Signal Name** (`eventB`) |
| `tests/manual/24-multiple-event/multiple-boundary.bpmn` | BoundaryEvent `multiBoundary` | `messageRef=msg1`, `<timerEventDefinition>PT10S` | Message Name, Correlation Key | **Timer Type / Timer Expression** |

Open each fixture, click the named element, confirm the listed missing fields are not present in the panel.

## Suggested fix

When the selected event has more than one event-definition child, render **one collapsible section per definition** in the property panel (Definition #1 — Message…, Definition #2 — Timer…), each with its existing variant-specific fields.

Source: BPMN-Editor property-panel gap audit. Cross-references regression test #27 (Multiple Event).

nightBaker:
## Summary

- Fixes #514: when a BPMN element has multiple event definitions (e.g., message + signal catch, message + timer boundary), all definitions are now visible and editable in the editor panel
- Single-definition elements are **completely unchanged** — no regression risk
- `bpmnEditor.js`: replaced the `break`-after-first extraction loop with a collect-all array + `forEach` alias scan; all existing flat fields (`hasTimerDefinition`, `messageName`, etc.) remain backward-compatible
- `ElementPropertiesPanel.razor`: existing single-def sections wrapped in `Count <= 1` gate; `Count > 1` shows labeled definition boxes using the same private fields and event handlers (editing routes to the first matching type — documented in the info bar)

## Key decisions

- **Backward-compatible aliases**: the `forEach` scan populates all flat fields from all definitions; last-write-wins for same-type multi-event (extremely rare in practice)
- **`@if/else if` chains** instead of `@switch` to avoid Razor markup-in-case issues
- **Activity Reference dropdown disabled** for Compensate in multi-def view (per design: `compensateActivities` not populated per-index)

## Test plan

- [x] `dotnet build Fleans.Web` — 0 errors
- [ ] Manual: open `24-multiple-event/message-or-signal-catch.bpmn` — verify Definition #1 (Message) and #2 (Signal) both show with editable fields
- [ ] Manual: open a single-signal catch — verify panel unchanged (no labeled boxes)
- See `tests/manual/53-multi-event-editor/test-plan.md` for full 5-step plan (requires Aspire + browser)

Closes #514


nightBaker (commented):
## Code Review — feat(#514): show all event definitions in editor for multi-event elements

### Summary

The implementation correctly refactors `bpmnEditor.js` from a break-after-first loop to a collect-all array + forEach alias scan, making all event definitions visible in the panel. The Razor component adds a clean multi-def branch with labeled definition boxes while keeping the single-def path fully backward-compatible.

### Findings

🟢 **Fixed during review:** Two pre-submit issues found and fixed before this review:
- `tests/manual/53-multi-event-editor/test-plan.md` was present in the git tree (delete was never staged in the merge commit) — removed via `git rm`, git recognised it as a rename.
- Test plan heading read `# Test 50` — corrected to `# Test 54`.

🟢 **JS correctness:** `entry.waitForCompletion = def.waitForCompletion !== false` correctly defaults `undefined` to `true`. The `else` branch that captures unknown `$type` strings prevents silent data loss. Flat backward-compat aliases are populated via `forEach` after the collect loop — last-write-wins for same-type multi-definitions, which is the documented limitation.

🟢 **Razor design:** Multi-def block binds each definition box to the same private fields as the single-def path. For a viewer scenario this is correct; for same-type multi-definitions the info bar explicitly warns that hand-editing XML is required. `compensateActivityRef` is shown as a read-only text field (not the dropdown) in multi-def view — this is intentional and noted in the info bar.

🟢 **`EventDefinitionData` class:** All fields have sensible defaults (`""` / `true`). Class is placed inside the `ElementPropertiesPanel` namespace, consistent with `EdgeData` and `CompensateActivity` in the same file.

PR #589, opened 2026-05-15
Shard streaming Stream IDs by WorkflowInstanceId (#565)

### LINKED ISSUE #565: Shard streaming Stream IDs by workflow instance to unblock parallel handlers
## Problem

`WorkflowEventsPublisher.cs` publishes every event of a given type to a **single fixed Stream Guid** derived from the type name:

```csharp
// src/Fleans/Fleans.Application/Events/WorkflowEventsPublisher.cs:65
StreamId.Create(WorkflowEventStreams.ExecuteCustomTaskStreamNamespace,
                nameof(ExecuteCustomTaskEvent));
```

Consequences for `ExecuteCustomTaskEvent`, `ExecuteScriptEvent`, `EvaluateConditionEvent`, `EvaluateActivationConditionEvent`, `EvaluateCompletionConditionEvent`, and the default `IDomainEvent` stream:

- All events of a type serialize through **one** subscriber grain activation (per implicit-subscription class).
- `TotalQueueCount = 8` in `FleanStreamingExtensions.cs:53` is wasted — only one of the 8 Redis stream queues ever receives traffic for a given event type (Stream Guid hashes to a single partition).
- Adding `fleans-worker` replicas does not increase parallelism — the single subscriber grain pins to one silo.

This is the throughput ceiling for `CustomTaskHandlerBase` plugins and all script/condition evaluation under load.

## Proposed fix

Shard the Stream Guid at publish time using `WorkflowInstanceId`:

```csharp
StreamId.Create(WorkflowEventStreams.ExecuteCustomTaskStreamNamespace,
                executeCustomTaskEvent.WorkflowInstanceId);
```

- **Per-workflow-instance ordering preserved** — events for the same instance stay FIFO on the same stream (matches Orleans per-stream ordering guarantees).
- **Cross-instance parallelism unlocked** — distinct instance IDs hash across all 8 Redis queues, distinct subscriber grain activations spread across silos via `WorkerPlacement`.

Apply uniformly to the 6 publish sites in `WorkflowEventsPublisher.OnPublish`:

| Event | Current key | Proposed key |
|---|---|---|
| `ExecuteCustomTaskEvent` | `nameof(ExecuteCustomTaskEvent)` | `event.WorkflowInstanceId` |
| `ExecuteScriptEvent` | `nameof(ExecuteScriptEvent)` | `event.WorkflowInstanceId` |
| `EvaluateConditionEvent` | `nameof(EvaluateConditionEvent)` | `event.WorkflowInstanceId` |
| `EvaluateActivationConditionEvent` | `nameof(EvaluateActivationConditionEvent)` | `event.WorkflowInstanceId` |
| `EvaluateCompletionConditionEvent` | `nameof(EvaluateCompletionConditionEvent)` | `event.WorkflowInstanceId` |
| Default `IDomainEvent` | `nameof(IDomainEvent)` | TBD (decide during analysis — may need per-event sharding) |

Subscribers (`CustomTaskHandlerBase`, `WorkflowExecuteScriptEventHandler`, condition evaluators, etc.) keep their `[ImplicitStreamSubscription]` attributes unchanged — they already accept any stream Guid in the namespace.

## Out of scope (separate follow-ups)

- Per-task-type namespace to eliminate the filter-after-deliver waste in `CustomTaskHandlerBase.OnNextAsync:77` (every plugin currently deserializes every event then drops mismatches).
- Bumping `TotalQueueCount` beyond 8 (only meaningful after this issue lands).
- Wiring real `CancellationToken` into plugin `ExecuteAsync` (`CustomTaskHandlerBase.cs:103` passes `CancellationToken.None`).
- Auditing `CompleteActivity`/`FailActivity` idempotency under stream redelivery.
- Batched publish via `OnNextBatchAsync` for bursty workflows.

## Acceptance

- All 6 publishers in `WorkflowEventsPublisher` use a sharded key.
- Existing `Fleans.Application.Tests` Orleans `TestCluster` runs stay green (per-stream ordering invariants intact).
- Manual regression: a workflow with two parallel multi-instance `serviceTask` branches shows handler grain activations on **different** silos when `fleans-worker` is scaled to 2+ replicas (verify via Orleans dashboard or activation-count log lines).
- CLAUDE.md updated: the "fixed stream Guid" trap and the per-instance sharding convention added to the streaming section.

## References

- Discovered while discussing `[ImplicitStreamSubscription]` parallelism — see `src/Fleans/Fleans.ServiceDefaults/FleanStreamingExtensions.cs:44` (Redis stream config) and `src/Fleans/Fleans.Worker/CustomTasks/CustomTaskHandlerBase.cs:26-73` (subscription wiring).
- Orleans docs on stream identity: per-stream FIFO is the only ordering guarantee; cross-stream is parallel.

nightBaker:
Closes #565.

## Summary

`WorkflowEventsPublisher` now keys each engine-event stream by `event.WorkflowInstanceId.ToString("D")` for the five non-default event types (`ExecuteCustomTaskEvent`, `ExecuteScriptEvent`, `EvaluateConditionEvent`, `EvaluateActivationConditionEvent`, `EvaluateCompletionConditionEvent`). Per-instance FIFO ordering is preserved (Orleans guarantees FIFO per stream); cross-instance traffic now distributes across `TotalQueueCount` Redis queues and across silos.

## Key changes

- **Publisher (`WorkflowEventsPublisher.cs`)**: five non-default `StreamId.Create` sites switched from `nameof(<Event>)` to `<event>.WorkflowInstanceId.ToString("D")`. Default `IDomainEvent` branch now logs `LogUnknownEventType` at Warning and drops, rather than fanning out to a dead-letter stream.
- **Subscribers** (5 files: `CustomTaskHandlerBase` + the four engine handlers): `OnActivateAsync` rebuilds stream id from `this.GetPrimaryKeyString()` (matching what Orleans implicit-subscription dispatch sets). The else-branch `SubscribeAsync` fallback was removed — implicit-subscription dispatch always materialises a handle, pinned by a new unit test.
- **`ILocalSiloDetails`** resolved via `this.ServiceProvider.GetRequiredService<>()` (service location) to avoid breaking `CustomTaskHandlerBase`'s public NuGet constructor signature.
- **Renamed** `WorfklowEvaluateConditionEventHandler` typo → `WorkflowEvaluateConditionEventHandler` (class + interface + both files). Dropped dead `Ping()` no-ops from activation-condition and completion-condition handlers.
- **Added** `Fleans.Application.Tests/Streaming/ImplicitSubscriptionHandleTest` — asserts `GetAllSubscriptionHandles().Count == 1` after one publish via `IEventPublisher`. `[DoNotParallelize]` for safety. Requires `InternalsVisibleTo Fleans.Application.Tests` on `Fleans.Application.csproj` (added).
- **Documentation**: CLAUDE.md gains two streaming bullets (sharding + subscriber-side trap). Both self-host guides gain a "Draining workflows for stream-format-changing upgrades" sub-section using `redis-cli` stream-length as the adapter-agnostic drain check. Manual regression entry 56 in `tests/manual/README.md` + new fixture folder `tests/manual/44-stream-sharding-parallelism/`.

## How it was reviewed

This PR is the implementation of issue #565's design that went through five rounds of analytic↔analysis-review (v1 → v5 + critiques in the issue comments). All four v1 blocking questions and the three v4 implementation gaps were resolved before this PR landed. v5 is the design that was approved → Ready.

## Test plan

- [x] `dotnet build` clean (0 errors, pre-existing test-analyzer warnings only)
- [x] `dotnet test Fleans.Application.Tests` — 319 passed, 6 expected skips, 0 failed (14m17s)
- [x] `dotnet test --filter "FullyQualifiedName!~Fleans.Application.Tests"` — 144 passed (Persistence) + 186 passed (Infrastructure), 0 failed
- [x] New `ImplicitSubscriptionHandleTest` passes in isolation
- [ ] Manual regression `tests/manual/44-stream-sharding-parallelism/test-plan.md` — pending QA run against a 2-replica `fleans-worker` deployment

## Notes for the reviewer

- The Helm drain doc uses `kubectl exec statefulset/fleans-redis -- redis-cli …` — the chart ships Redis as a StatefulSet (verified). Both self-host guides include a "verify the actual key pattern" step before running the drain loop.
- Two minor follow-ups from v5 review (consolidating the per-class hook into a single shared static, and the literal `deployment/` → `statefulset/` correction the implementer should verify against their own cluster) are noted in the v5 review comment on the issue and don't gate merge.


nightBaker (commented):
-

nightBaker (commented):
## Code Review — Shard streaming Stream IDs by WorkflowInstanceId (#565)

### Scope verification
- Diff against the v5 design (issue #565): all 5 phases land in one PR as specified.
  - Phase 1 ✓ (5 subscribers rewritten, class+interface+file rename for the `Worfklow…` typo, dead `Ping()` no-ops dropped, `InternalsVisibleTo` added).
  - Phase 2 ✓ (5 publisher sites switched to `WorkflowInstanceId.ToString("D")`, default branch logs + drops).
  - Phase 3 ✓ (`ImplicitSubscriptionHandleTest` pinning `GetAllSubscriptionHandles().Count == 1`).
  - Phase 4 ✓ (CLAUDE.md 2 streaming bullets, both self-host guides gain drain sub-section).
  - Phase 5 ✓ (`tests/manual/44-stream-sharding-parallelism/` + README entry 56).
- Test results from the implementation phase: 319 passed in `Fleans.Application.Tests` (incl. the new test), 144 passed in Persistence, 186 passed in Infrastructure. No regressions.

### CLAUDE.md compliance
- **`[LoggerMessage]` source generators** (CLAUDE.md "Logging" rule) — every new log point (5 × `LogActivated` + `LogUnknownEventType`) uses `[LoggerMessage]`. ✓
- **`List<T>` for Orleans-serialized fields** — not touched. ✓
- **Stream-id sharding + subscriber-side trap** (the very bullets this PR adds to CLAUDE.md) — code matches the documented contract. ✓

### 🟡 Important

1. **Test side-effect noise.** `ImplicitSubscriptionHandleTest` publishes an `ExecuteScriptEvent` with `WorkflowInstanceId: Guid.NewGuid()`. The subscriber's `OnNextAsync` will then try to `GetVariables` on an unbacked `IWorkflowInstanceGrain` and the script-execution path will throw. The TCS is set before any of that fires (good — the assertion passes correctly), but the test still leaves error log lines and routes through `FailActivity` against a non-existent instance. Functionally harmless, but a follow-up that resolves the activation hook *without* triggering downstream `OnNextAsync` work would be cleaner. Not blocking.

### 🟢 Minor

1. **Pre-existing EventId duplicates.** This PR's `LogActivated` IDs (4006, 4015, 4026, 4035, 4036) are all new and don't collide within a single source class. The codebase already has pre-existing duplicates of 4030–4034 between `CustomTaskHandlerBase` and `WorkflowEvaluateCompletionConditionEventHandler` — out of scope, but worth filing a chore issue eventually.
2. **Drain commands use literal container/resource names.** Compose: `docker exec orleans-redis …` — name is fixed by `postprocess-compose-bundle.sh`, so OK. Helm: `kubectl exec statefulset/fleans-redis …` — depends on the chart release name (default "fleans" → matches). Both guides explicitly tell operators to verify the actual key pattern via `KEYS '*'` first, which is the right discipline.
3. **`OnImplicitActivation` static hook is null in production.** The cost is one null-check per grain activation — acceptable, but if test parallelism is ever introduced (the test is correctly `[DoNotParallelize]`d), the hook is process-static. Future cleanup could consolidate the per-subscriber hooks into a single shared `Action<Type,int>?` — not a blocker.

### Verdict

No 🔴 Blocking issues. The substance matches the v5 design that survived 5 rounds of analysis-review. CI rollup is currently PENDING — once it lands green, the worker can promote to Testing.

OK to promote.

src/Fleans/Fleans.Application/Events/Handlers/WorkflowEvaluateActivationConditionEventHandler.cs
```diff
@@ -26,22 +27,23 @@ public WorkflowEvaluateActivationConditionEventHandler(
         _grainFactory = grainFactory;
     }
 
+    // Test-only hook fired after handles are resumed. Null in production.
+    internal static Action<int>? OnImplicitActivation;
```
nightBaker: we don't have to change code only for testing purpose

website/src/content/docs/guides/self-host-docker-compose.md
```diff
@@ -177,6 +177,38 @@ For breaking version bumps (e.g. major release), read the GitHub Release
 notes — they call out any migration steps Compose itself does not perform
 (secret rotations, value renames, removed env vars).
 
+### 6.1. Draining workflows for stream-format-changing upgrades
+
+Releases flagged as **"stream upgrade"** in the release notes change Orleans
+stream identity — messages on the old stream key become orphaned after the
+upgrade, so in-flight workflows must drain first. Procedure (no engine
+changes required, works for any Redis-streaming adapter version):
```
nightBaker: does it work only for redis streaming ?

website/src/content/docs/guides/self-host-docker-compose.md
```diff
@@ -177,6 +177,38 @@ For breaking version bumps (e.g. major release), read the GitHub Release
...
+
+```bash
+# 1. Stop API/Web/Mcp so no new work arrives. Workers keep running and
+#    continue to consume queued events.
+docker compose stop fleans-api fleans-web fleans-mcp
```
nightBaker: we don't need this section because we still don't have first version of engine, so we can add breaking changes

nightBaker:
do not add only test purpose code in source code
do not add info how to move from old versions to new till we get to v1.0.0 (we can add breaking changes)

src/Fleans/Fleans.Application/Events/Handlers/WorkflowEvaluateActivationConditionEventHandler.cs
nightBaker: Agreed — removed `OnImplicitActivation` from all 5 production handler grains (script, condition, activation-condition, completion-condition, custom-task) and deleted `ImplicitSubscriptionHandleTest` along with the now-unused `InternalsVisibleTo` on `Fleans.Application.csproj`.

The implicit-subscription dispatch path is exercised end-to-end by every script/condition/custom-task integration test (CancelEventTests, CompensationEventTests, ConditionalEventTests, EventSubProcess*Tests, MessageStartEventTests, MultiInstanceTests, SignalStartEventTests, …). If implicit-subscription dispatch regresses, dozens of those fail at once — the regression signal is already there without a dedicated unit test poking holes in production code.

Also updated the CLAUDE.md "subscriber-side stream-id trap" entry to point at the broader integration suite instead of the deleted unit test.

website/src/content/docs/guides/self-host-docker-compose.md
nightBaker: Moot — the whole "Draining workflows" section is gone (see thread below). Pre-v1, breaking changes are allowed without a formal drain procedure.

website/src/content/docs/guides/self-host-docker-compose.md
nightBaker: Agreed. Removed §6.1 in `self-host-docker-compose.md` and §5.1 in `self-host-helm.md` — both new sections introduced by this PR. Pre-v1 engine, breaking changes are fine without a formal drain procedure.

nightBaker:
Addressed all 3 review threads:

1. **`OnImplicitActivation` test-only hook** — removed the static field + invocation from all 5 implicit-subscription handler grains (`WorkflowExecuteScriptEventHandler`, `WorkflowEvaluateActivationConditionEventHandler`, `WorkflowEvaluateConditionEventHandler`, `WorkflowEvaluateCompletionConditionEventHandler`, `CustomTaskHandlerBase`). Deleted `Fleans.Application.Tests/Streaming/ImplicitSubscriptionHandleTest.cs` and the now-unused `InternalsVisibleTo` entry on `Fleans.Application.csproj`. Updated the CLAUDE.md "subscriber-side stream-id trap" entry to drop the cross-reference to the deleted test. The implicit-subscription dispatch path is still covered end-to-end by the broader script/condition/custom-task integration suite — dozens of tests fail simultaneously if it regresses.

2. **`self-host-docker-compose.md` "Redis only?"** — moot; whole section removed (see #3).

3. **`self-host-docker-compose.md` / `self-host-helm.md` "drain workflows" section** — deleted §6.1 (compose guide) and §5.1 (helm guide). Pre-v1 engine, breaking changes are allowed without a formal drain procedure.

Build + full `Fleans.Application.Tests` suite green after the changes.

nightBaker (commented):
## Code Review — refactor(#565): address review feedback (commit `db4fb6e`)

Re-review pass on the new commit that addresses the 3 unresolved threads from the prior round.

### Scope verification

Diff: 1 insertion / 150 deletions across 10 files, all directly mapping to the three threads:

- **Thread 1 (`OnImplicitActivation` test-only hook)** — removed the `internal static Action<int>? OnImplicitActivation` field + its `?.Invoke(handles.Count)` call from all 5 `[ImplicitStreamSubscription]` handler grains (`WorkflowExecuteScriptEventHandler`, `WorkflowEvaluateActivationConditionEventHandler`, `WorkflowEvaluateConditionEventHandler`, `WorkflowEvaluateCompletionConditionEventHandler`, `CustomTaskHandlerBase`). `Fleans.Application.Tests/Streaming/ImplicitSubscriptionHandleTest.cs` deleted. `InternalsVisibleTo Include="Fleans.Application.Tests"` dropped from `Fleans.Application.csproj` — verified no other internal members in `Fleans.Application` are consumed by tests (only `CorePlacementStrategy.Singleton` / `CorePlacementDirector.HasCoreRole` exist and are unused from test code; Worker's `InternalsVisibleTo` stays because tests do reference `WorkerPlacementStrategy.Singleton` / `WorkerPlacementDirector`). CLAUDE.md "subscriber-side stream-id trap" cross-reference rewritten to point at the broader script/condition/custom-task integration suite instead of the deleted test.
- **Threads 2 & 3 (drain-workflows docs)** — §6.1 in `website/src/content/docs/guides/self-host-docker-compose.md` and §5.1 in `website/src/content/docs/guides/self-host-helm.md` deleted in full. Pre-v1, breaking changes are allowed without a formal drain procedure, so the section had no current audience.

### Risks

- **Lost coverage signal for "exactly one implicit-subscription handle":** the deleted test was the only direct pin on that invariant. The CLAUDE.md note now explicitly delegates the regression signal to the broader integration suite — if implicit-subscription dispatch produces 0 handles, every script/condition/custom-task end-to-end test loses its trigger and fails; if it produces ≥2, completion callbacks fire multiple times and the `WorkflowInstance` aggregate's idempotency guards either swallow or surface duplicates. Realistic regression signal stays intact, just transitively.
- **No production-code surface area lost:** the removed `OnImplicitActivation` field is a static `Action<int>?` only ever invoked once per activation with `?.` short-circuit — its removal is functionally a no-op at runtime.

### CLAUDE.md compliance

- ✅ "WorkflowInstance state changes flow through `DrainAndRaiseEvents()`" — handler grains were not touched in a way that affects state-mutation paths.
- ✅ "Log all workflow instance state changes" — `LogActivated` (the only log call in the touched `OnActivateAsync` methods) is preserved.
- ✅ "Logging: always use `[LoggerMessage]` source generators" — no new log sites added.
- ✅ Production code no longer carries test-only members.

### Verification

- `dotnet build` clean (0 errors, pre-existing warnings unchanged).
- Targeted suites green: `EventPublisherTests`, `ConditionalEventTests`, `ConditionalStartEventTests`, `MultiInstanceTests`, `MessageStartEventTests` (22 tests, 0 failures) — these exercise all 4 affected handler dispatch paths through implicit subscriptions, validating the "test deletion didn't lose the invariant" claim.

PR #594, opened 2026-05-16
feat(#566): per-task-type stream namespace for custom-task handlers

### LINKED ISSUE #566: Per-task-type stream namespace for custom-task handlers (eliminate filter-after-deliver waste)
## Problem

All custom-task plugins subscribe to a single shared stream namespace:

```csharp
// src/Fleans/Fleans.Worker/CustomTasks/CustomTaskHandlerBase.cs:26
[ImplicitStreamSubscription(WorkflowEventStreams.ExecuteCustomTaskStreamNamespace)]
```

Each handler then filters by `TaskType` and drops mismatches:

```csharp
// src/Fleans/Fleans.Worker/CustomTasks/CustomTaskHandlerBase.cs:77
if (!string.Equals(item.TaskType, TaskType, StringComparison.OrdinalIgnoreCase))
    return;
```

With **N** registered plugin types and **M** custom-task events, every silo:
- Activates **N** subscriber grains per stream (one per plugin class).
- Deserializes **N×M** messages from Redis.
- Discards **(N−1)×M** of them after the type check.

This is wasted CPU, GC pressure, and Redis bandwidth that grows linearly with the plugin catalog. Today the cost is modest (`RestCallerHandler` only). The fanout becomes painful once a deployment carries 5–10+ plugins.

## Proposed fix

Make the task type part of the stream namespace, not a filter:

```csharp
// publish (in WorkflowEventsPublisher)
var ns = $"{WorkflowEventStreams.ExecuteCustomTaskStreamNamespace}.{executeCustomTaskEvent.TaskType}";
StreamId.Create(ns, executeCustomTaskEvent.WorkflowInstanceId);

// subscribe (each plugin declares its own namespace)
public abstract class CustomTaskHandlerBase
{
    // Subclasses override; attribute applied via subclass:
    //   [ImplicitStreamSubscription("events.ExecuteCustomTaskEvent.rest-call")]
    //   public sealed class RestCallerHandler : CustomTaskHandlerBase { ... }
}
```

Result: each plugin's grain only sees its own traffic. The `TaskType` filter in `OnNextAsync` becomes a no-op (or an assertion).

### Wiring decision: how subclasses declare their namespace

Three options to evaluate during analysis:

1. **Hard-coded `[ImplicitStreamSubscription]` per subclass.** Simplest; plugin author writes the attribute alongside `TaskType`. Duplication risk if the two strings drift.
2. **Source generator** that emits the attribute from `TaskType`. Eliminates drift; adds a generator project.
3. **Namespace pattern via Orleans' `ImplicitStreamSubscriberTable` customization.** Most flexible; deepest Orleans coupling.

Recommend option 1 in the design — plugin authors already declare `TaskType`; adding a matching attribute string is no worse than declaring the type itself, and the build will fail loudly if they forget.

## Dependency on #565

#565 shards Stream IDs by `WorkflowInstanceId` within a namespace. This issue partitions the **namespace** by `TaskType`. The two are orthogonal but compose:

- After #565 alone: one namespace, N×M deliveries, but parallel across instances within each plugin grain.
- After both: N namespaces, M/N average deliveries per plugin, parallel across instances within each.

#565 should land first so the publish-side shard key is already in place when this change extends the namespace string.

## Out of scope

- Same OOS list as #565 (cancellation token, idempotency audit, batched publish, `NumQueues` bump).
- Extending this pattern to `ExecuteScriptEvent` / `EvaluateConditionEvent` etc. — those subscribers don't filter by a discriminator, so they don't have this problem.

## Acceptance

- `WorkflowEventsPublisher` publishes `ExecuteCustomTaskEvent` to `events.ExecuteCustomTaskEvent.{TaskType}`.
- `CustomTaskHandlerBase` subclasses declare their per-type `[ImplicitStreamSubscription]` (mechanism per design doc).
- The `TaskType` mismatch branch in `OnNextAsync` becomes unreachable under normal operation (keep it as a defensive log + drop).
- Manual regression: with two plugins (`RestCallerHandler` + a stub second handler) registered on the same silo, only the matching grain class activates for a given task type — verified via Orleans dashboard or activation logs.
- CLAUDE.md "Plugin-author NuGet stack" / custom-tasks section updated with the per-type namespace convention so plugin authors know what attribute to write.
- `Fleans.Plugins.RestCaller` updated to match.
- Existing tests green.

## References

- Discovered alongside #565 during `[ImplicitStreamSubscription]` parallelism review.
- `src/Fleans/Fleans.Worker/CustomTasks/CustomTaskHandlerBase.cs:26-77` (subscription + filter)
- `src/Fleans/Fleans.Application/Events/WorkflowEventsPublisher.cs:63-66` (publish site)
- `src/Fleans/Fleans.Application.Abstractions/Events/WorkflowEventStreams.cs:33` (namespace constant)

nightBaker:
Closes #566.

## Summary

Implements the per-task-type stream namespace for `ExecuteCustomTaskEvent`. After this change, each plugin's `CustomTaskHandlerBase` subclass receives only its own traffic — eliminates the N-plugins × M-events filter-after-deliver fanout that grows linearly with the plugin catalog. Approved Design & Plan v3 in the issue thread (3 revisions through analysis-review).

## What changed

- **`WorkflowEventStreams.GetExecuteCustomTaskNamespace(string taskType)`** helper added (`events.ExecuteCustomTaskEvent.{taskType}`). Publisher and base handler's `OnActivateAsync` route through it.
- **`WorkflowEventsPublisher.Publish(ExecuteCustomTaskEvent)`** now publishes to the per-type namespace.
- **`CustomTaskHandlerBase`**: class-level `[ImplicitStreamSubscription]` dropped (Orleans only walks the concrete grain class — base attribute was never honored). `OnActivateAsync` resumes via the per-type namespace; `OnNextAsync` mismatch path logs a warning (`LogTaskTypeMismatch`, EventId 4037) instead of silently returning — the branch is unreachable under correct routing.
- **`AddCustomTaskPlugin<T>(taskType, …)`** validates at silo startup and throws `InvalidOperationException` on:
  - (a) duplicate `TaskType` registration across handlers;
  - (b) `typeof(T)` missing `[ImplicitStreamSubscription("events.ExecuteCustomTaskEvent.<taskType>")]` matching the helper. Attribute discovery uses `CustomAttributeData` (byte-equivalent to what Orleans itself reads during grain-class discovery).
- **`RestCallerHandler`**: attribute literal switched to `"events.ExecuteCustomTaskEvent.rest-call"`.
- **`Fleans.Application.Tests/CustomTasks/CustomTaskRoutingTests.cs`** (3 tests, all pass):
  1. `Per_type_namespace_is_used_when_publishing_custom_task_event` — subscribes test observers to two per-type streams (`test-no-op-a`, `test-no-op-b`); publishes an event for type A through the publisher; asserts only the A-stream observer fired.
  2. `AddCustomTaskPlugin_throws_on_duplicate_TaskType` — direct unit test on the validator's duplicate path.
  3. `AddCustomTaskPlugin_throws_when_attribute_namespace_drifts_from_TaskType` — uses the `BadAttributeCustomTaskHandler` stub (drifted attribute string) to verify the attribute-correctness path.
- **Docs**: `CLAUDE.md` gains a new "Custom-task per-type stream namespace" code-conventions bullet + the "How to Add a Custom Task Plugin" step 4 is extended with the registration-time validation contract. `website/src/content/docs/concepts/custom-tasks.md` updates the "How it works" steps 2-3 to describe per-type routing, notes the literal-string `[ImplicitStreamSubscription]` requirement under "Authoring a plugin", and adds an "Upgrade note: per-type stream namespaces" section calling out (i) in-flight-event orphaning across the rollout (pre-v1; no drain procedure) and (ii) plugin authors using the `fleans-custom-worker-example` template need to add the per-type attribute. Removes the now-stale "Per-task-type stream partitioning is deferred" limitation.

## Test plan

- [x] `dotnet build` clean from `src/Fleans/`.
- [x] `dotnet test --filter "FullyQualifiedName~CustomTaskRoutingTests"` — 3/3 passed.
- [ ] Full `Fleans.Application.Tests` suite green (run by CI).
- [ ] Manual regression suite — to be assessed at testing time; this PR is runtime-affecting (publisher + base handler stream-id construction) so warrants the BPMN regression list.

## Follow-up (separate issue)

While verifying EventId 4036 wasn't free (it is in use by the completion-condition handler's `LogActivated` added in #565), this PR's grep surfaced a pre-existing **double-assignment of EventIds 4030–4034** between `WorkflowEvaluateCompletionConditionEventHandler.cs` and `CustomTaskHandlerBase.cs`. Out of scope here — should be filed as a separate hygiene issue and resolved by renumbering one of the families (most likely the custom-task family to 4040+). The new `LogTaskTypeMismatch` at EventId 4037 sits inside the family that will need to move; the follow-up should pick it up too.


nightBaker (commented):
## Code Review — feat(#566): per-task-type stream namespace for custom-task handlers

Solid PR. Design has been through 3 analysis-review rounds, source-level changes match the approved Design & Plan v3, validation contract is well-tested, and the docs surface (CLAUDE.md + concepts/custom-tasks.md upgrade note) is complete. No 🔴 blocking issues. OK to promote.

### What I verified

- **`WorkflowEventStreams.GetExecuteCustomTaskNamespace`** centralises the literal `events.ExecuteCustomTaskEvent.{taskType}` and is referenced from both sides (publisher in `WorkflowEventsPublisher.cs:64+` and subscriber in `CustomTaskHandlerBase.OnActivateAsync`), so drift can only come from the `[ImplicitStreamSubscription]` attribute literal — which is precisely what `AddCustomTaskPlugin<T>` validates at registration via `CustomAttributeData` (byte-equivalent to what Orleans reads during grain-class discovery). Good defense.
- **Subscriber-side stream-id reconstruction** in `CustomTaskHandlerBase.OnActivateAsync` still rebuilds `streamId` from `this.GetPrimaryKeyString()` — preserves the CLAUDE.md "Subscriber-side stream-id trap" invariant. The namespace component switched from constant to per-`TaskType`, but the key derivation is unchanged. ✅
- **`OnNextAsync` mismatch path** now logs (`LogTaskTypeMismatch`, EventId 4037) instead of silently returning. Correct: the branch is unreachable under per-type routing, so surfacing future drift is the right move.
- **`[ImplicitStreamSubscription]` removed from `CustomTaskHandlerBase` (abstract base)** — correct, since Orleans only walks concrete grain classes for `[ImplicitStreamSubscription]` (inherit:false), and a base-class attribute would be misleading.
- **Validation tests** cover both paths (duplicate `TaskType` + drifted attribute) and use `ThrowsExactly<InvalidOperationException>` with explicit message asserts. Routing test uses a per-stream `CountingObserver` and asserts `streamB.Count == 0` to verify the fanout-elimination invariant directly.
- **`RestCallerHandler`** attribute literal `"events.ExecuteCustomTaskEvent.rest-call"` matches `GetExecuteCustomTaskNamespace("rest-call")`. ✅ — and `AddRestCallerPlugin` (the supported registration entry point) will catch any future drift.
- **Docs:** new CLAUDE.md bullet documents the contract; `concepts/custom-tasks.md` adds the "Upgrade note: per-type stream namespaces" section that explicitly calls out (a) in-flight event orphaning across the rollout (pre-v1, no drain) and (b) the literal-string attribute requirement plus the fanout-elimination guarantee.

### 🟡 Important (follow-ups, not blocking)

1. **EventId 4037 sits inside an already-double-assigned family (4030–4034).** I confirmed via grep that `WorkflowEvaluateCompletionConditionEventHandler.cs:96-112` and `CustomTaskHandlerBase.cs:137-153` both occupy IDs 4030, 4031, 4032, 4033, 4034 — they're in the same logical family per `docs/plans/2026-02-08-structured-workflow-logging.md` (Event handlers: 4000-4099) but neither has a documented sub-range. Adding 4037 to this family doesn't itself create a new collision, but it grows the renumbering surface for the cleanup the PR body alludes to. Suggest:
   - File the hygiene issue now (single `gh issue create` referencing this PR), so the follow-up isn't lost.
   - Pick a sub-range convention in the same `docs/plans/2026-02-08-structured-workflow-logging.md` table when the renumber lands — e.g., reserve 4030-4039 for completion-condition and move custom-task to 4040-4049 (which also picks up the new 4037 → 4047).

2. **Validation only fires through `AddCustomTaskPlugin<T>`.** An external host that bypasses this helper (e.g., registering `CustomTaskPluginDescriptor` directly, or skipping descriptor registration entirely) would not be validated at startup and could silently misroute. The comment at `CustomTaskServiceCollectionExtensions.cs` "relies on AddSingleton(new …) (instance-based registration) — the supported entry point" already names this. Given that CLAUDE.md and the new docs both reinforce "must call `AddCustomTaskPlugin<T>`", this is acceptable, but worth keeping in mind if/when an alternate registration shape ships.

### 🟢 Minor

- The two `TestNoOpCustomTaskHandlerA/B` stubs expose `public static int InvocationCount;` and `Interlocked.Increment` on it inside `ExecuteAsync`, but the routing test only reads `CountingObserver.Count`. The static-int fields are dead today. Either drop them or wire a second assertion (`Assert.AreEqual(0, TestNoOpCustomTaskHandlerB.InvocationCount)`) so the stub surface stays meaningful.
- `CustomTaskRoutingTests.Per_type_namespace_is_used_when_publishing_custom_task_event` polls observerA for up to 5 s but asserts on observerB immediately after. The asymmetry is fine in practice (B should never receive A's events under any condition) but worth a one-line comment so future readers don't add a "wait for B to also drain" loop and reintroduce flakiness.
- The `EventId 4037` log message wraps across two strings (`"...handler claims '{HandlerTaskType}' — dropping. " + "Indicates publisher/attribute drift..."`). `[LoggerMessage]` source generator accepts a single string template; concatenation works but slightly obscures the message at the call site. Cosmetic.

| Rule | Verdict |
|---|---|
| `[LoggerMessage]` source-generator for all new logging | ✅ `LogTaskTypeMismatch` is `[LoggerMessage] private partial void` |
| Documentation in same PR | ✅ CLAUDE.md bullet + `concepts/custom-tasks.md` updated |
| Records / `[GenerateSerializer]` for serialized types | N/A — no new DTOs |
| `Fleans.Worker` leaf-package boundary preserved | ✅ — `CustomTaskHandlerBase` only references `Fleans.Application.Abstractions` (`WorkflowEventStreams`) and `Fleans.Domain.Abstractions` (event records); no `Fleans.Application` / `Fleans.Domain` / `Fleans.Infrastructure` references introduced |
| Subscriber-side stream-id trap (key derived from `GetPrimaryKeyString()`) | ✅ preserved |

— hourly-code-review

PR #596, opened 2026-05-16
feat(#568): grain-lifetime CancellationToken in CustomTaskHandlerBase

nightBaker:
Closes #568.

## Summary

`CustomTaskHandlerBase.ExecuteAsync` now receives a `CancellationToken` whose source is the handler grain's lifetime (cancelled in a new `OnDeactivateAsync` override). When Orleans deactivates the grain (silo scale-down, idle collection, shutdown), the token is signalled; the in-flight plugin should propagate `OperationCanceledException`, and the base class catches that path via a `when` filter, re-throws, and lets the stream provider redeliver the event after reactivation — no `FailActivity` call.

Plugin-internal cancellation (a plugin's own timeout firing) still routes to `FailActivity` — the `when (_grainLifetimeCts?.IsCancellationRequested == true)` filter distinguishes the two cases.

Implementation matches the v3 design from analysis-review with the two 🟡 wording fixes from round-2 review applied (no "lose the in-flight event" wording; Orleans 10.x not 9.2.1; concrete manual-test plan number #58; risk-#2 mechanism rationale updated to reference Orleans turn-based scheduling).

## What changed

- `Fleans.Worker/CustomTasks/CustomTaskHandlerBase.cs` (+30):
  - New `_grainLifetimeCts` field with `// Cancel-only — no Timer, no unmanaged handles` comment.
  - `OnActivateAsync` initialises it.
  - **New** `OnDeactivateAsync` override cancels it.
  - `OnNextAsync` threads `_grainLifetimeCts!.Token` into `ExecuteAsync` (replaces `CancellationToken.None` at the call site).
  - **New** `catch (OperationCanceledException) when (_grainLifetimeCts?.IsCancellationRequested == true)` before the generic catch, with rationale comment.
  - **New** `[LoggerMessage]` at EventId 4050 — picked outside the known 4030-4034 family that pre-dates the EventId-hygiene follow-up to #566.
- `Fleans.Plugins.RestCaller.Tests/RestCallerHandlerCancellationTests.cs` (new, 2 tests):
  - `ExecuteAsync_PreCancelledToken_PropagatesOperationCanceledException` — happy-path token threading.
  - `ExecuteAsync_TokenCancelledMidRequest_PropagatesOperationCanceledException` — mid-flight cancellation via the `BlockingHandler` stub.
- `website/src/content/docs/concepts/custom-tasks.md` — new "Cancellation" section with the "redeliver, not lose" wording.
- `CLAUDE.md` "How to Add a Custom Task Plugin" step 2 — extended with the OCE exception and `when`-filter discriminator reference.
- `tests/manual/58-custom-task-cancellation/test-plan.md` (new, 4-step plan):
  - Steps 1-3: the deactivation-mid-call → redelivery path.
  - Step 4: plugin-internal timeout still routes to FailActivity (the `when` filter discriminator validates here).

## Behavior change

Plugins that already thread their input `CancellationToken` correctly (RestCallerHandler's linked-CTS pattern at `RestCallerHandler.cs:113`) get cleaner silo deactivation for free.

Plugins that ignore the token block silo deactivation until Orleans's hard timeout expires (~30 s); the stream record is **not lost**, redelivery still happens on the new activation. The cost is silo-shutdown latency, not correctness.

## Deferred

The integration test verifying the base-class OCE catch path end-to-end requires `WorkflowTestBase` to expose a custom-task-plugin registration hook, which it currently doesn't. Manual regression #58 covers the end-to-end contract (Path A: Aspire dashboard stop/start; Path B: process-level kill). The base-class catch logic itself is small enough to verify by code review; the unit test in `Fleans.Plugins.RestCaller.Tests` pins the plugin-side token-threading contract that the catch depends on.

## Dependency note

This PR's edits hit the same `OnNextAsync` try/catch body that PR #594 (for #566, currently in Review by Human) restructures with the per-task-type stream namespace work. The two compose cleanly in either merge order:
- If #594 merges first, the catch block addition here drops in alongside the new `LogTaskTypeMismatch` warning catch already added by #594.
- If this PR merges first, #594 will need to re-base its OCE-mismatch catch around our new OCE-as-cancellation catch (also trivial).

No semantic dependency in either direction.

## Test plan

- [x] `dotnet build` from `src/Fleans/` — 0 errors.
- [x] `dotnet test --no-build Fleans.Plugins.RestCaller.Tests` — 18 / 0 / 0 (16 pre-existing + 2 new).
- [x] `dotnet test --no-build Fleans.Application.Tests --filter "FullyQualifiedName~CustomTask|FullyQualifiedName~ScriptTask|FullyQualifiedName~TaskActivity"` — 18 / 0 / 0, no regression on the custom-task / script-task / task-activity surface area.
- [x] `npm run build` from `website/` — 31 pages built clean.
- [ ] Full `Fleans.sln` test suite green (run by CI).
- [ ] Manual regression `tests/manual/58-custom-task-cancellation/` (4-step plan) — deferred to manual-regression-testing skill.


nightBaker (commented):
## Code Review — feat(#568): grain-lifetime CancellationToken in CustomTaskHandlerBase

Tight, well-scoped PR. Implementation matches the v3 design from analysis-review and the two 🟡 wording fixes from round-2 review are both applied (no "lose the in-flight event"; Orleans 10.x not 9.2.1; concrete manual-test number #58). Inline comments explain the non-obvious bits (Orleans turn-based scheduling rationale on the `when` filter; CTS Dispose elided because Cancel-only). No blocking issues.

### What I verified

- **`CustomTaskHandlerBase.cs:34-36` (new CTS field):** `private CancellationTokenSource? _grainLifetimeCts;` with inline comment naming the no-Timer/no-Dispose invariant. Initialised in `OnActivateAsync` at L60 — before `await base.OnActivateAsync(...)` runs, so any base-class subscriber sees a non-null token field. ✅
- **`OnDeactivateAsync` (new override, L79-86):** `_grainLifetimeCts?.Cancel()` then `await base.OnDeactivateAsync(...)`. Null-safe; correct order (child override cancels first, then base lifecycle runs). ✅
- **`OnNextAsync` (L116):** `_grainLifetimeCts!.Token` replaces `CancellationToken.None`. Null-forgiving safe — Orleans grain lifecycle guarantees `OnActivateAsync` runs before any `OnNextAsync`. ✅
- **New OCE catch (L125-132):** `catch (OperationCanceledException) when (_grainLifetimeCts?.IsCancellationRequested == true)` placed BEFORE the generic `catch (Exception)`. C# clause-order semantics — generic catch would otherwise swallow OCE (IS-A Exception). The null-coalesce-to-false (`?.` returning null compared to `true` → false) means if the field were somehow null, the OCE falls through to the regular failure path. Defensive against impossible-but-not-statically-impossible states. ✅
- **`LogPluginCancelledOnDeactivation` (L183-186):** EventId 4050, deliberately picked outside the known 4030-4034 family (per round-1 minor). Source-generator template references `{TaskType}` and `{ActivityId}` — `[LoggerMessage]` matches by name (not position), so the parameter list `(string activityId, string taskType)` binds correctly regardless of placeholder order. ✅
- **Tests** (`Fleans.Plugins.RestCaller.Tests/RestCallerHandlerCancellationTests.cs`): 2 new — pre-cancelled token and mid-request cancellation paths. Uses the existing `ExecuteForTest` seam. Built locally, both pass alongside the 16 existing tests. ✅
- **Docs:** `concepts/custom-tasks.md` "Cancellation" section uses the round-2 review's "redeliver, not lose" wording. CLAUDE.md step 2 names `_grainLifetimeCts` and the `when` filter discriminator, so the next contributor can grep into the production code. ✅
- **Manual regression** `tests/manual/58-custom-task-cancellation/test-plan.md`: 4-step plan — Step 4 specifically contrasts plugin-internal timeout (FailActivity) vs deactivation cancellation (redelivery), which is the discriminator the `when` filter is supposed to enforce. Good test design. ✅

| Rule | Verdict |
|---|---|
| `[LoggerMessage]` source generators for all logging | ✅ `LogPluginCancelledOnDeactivation` is `[LoggerMessage] private partial void` |
| EventId range discipline (4000-4099 = event handlers) | ✅ EventId 4050 is inside the range and outside the known 4030-4034 double-assignment family |
| Documentation in same PR | ✅ CLAUDE.md, `concepts/custom-tasks.md`, manual regression #58 all included |
| Records with `[GenerateSerializer]` | N/A — no new DTOs |
| `Fleans.Worker` leaf-package boundary | ✅ — no new references to `Fleans.Application` or `Fleans.Domain` (still only `Fleans.Application.Abstractions` + `Fleans.Domain.Abstractions` via existing usings) |

### 🟢 Minor

1. **Test coverage for the base-class catch.** The PR description acknowledges this and points to manual regression #58. Worth keeping in mind: if a future change to the `when` filter breaks the discriminator (e.g., someone refactors `_grainLifetimeCts` to a property or moves the Cancel() call out of `OnDeactivateAsync`), no unit test would catch it. The current setup is defensible given the `WorkflowTestBase` plugin-registration gap, but a follow-up issue tracking "wire `AddCustomTaskPlugin<T>` into `WorkflowTestBase` so the OCE catch can have an integration test" would prevent the gap from growing. Not blocking.

2. **`_grainLifetimeCts` is reused across reactivations.** Orleans can reactivate a grain after deactivation (e.g., a new event arrives). On reactivation, `OnActivateAsync` runs again, replacing the old (cancelled) CTS with a fresh one — old reference is unreferenced and GC'd. Behaviour is correct, but worth a one-line code comment near the field declaration noting the reactivation contract: *"Replaced (not reused) on each `OnActivateAsync` — the old cancelled CTS becomes unreferenced and GC'd."* Helps future readers reason about the lifecycle.

3. **`OnActivateAsync` initialises `_grainLifetimeCts` after the field assignment but inside the override.** If the activation flow ever needs to subscribe to a stream callback that itself observes cancellation (no such callback today), the assignment should be the *very first* statement to make that future-safe. It already is — line 60. Mentioned only for the record; no change needed.

— hourly-code-review

PR #597, opened 2026-05-16
docs(#569): pin stream-redelivery idempotency contract

### LINKED ISSUE #569: Audit CompleteActivity/FailActivity idempotency under stream redelivery
## Problem

`src/Fleans/Fleans.Worker/CustomTasks/CustomTaskHandlerBase.cs:121-123` deliberately re-throws if `FailActivity` itself fails, relying on Orleans stream redelivery + a comment that says "domain idempotency guards handle duplicates":

```csharp
catch (Exception failEx)
{
    LogFailActivityFailed(failEx, item.ActivityId);
    throw; // let stream provider retry — domain idempotency guards handle duplicates
}
```

This claim is **load-bearing for correctness** but unverified at the audit level. Once #565/#566/#567 land and throughput goes up, redelivery scenarios become orders of magnitude more frequent and any latent non-idempotency surfaces as silent data corruption (duplicated outputs merged twice, completed activities re-failed, compensation double-fired).

The same redelivery concern applies to `CompleteActivity` even though it's not in the catch path — any transient RPC error inside the call could land state on the workflow instance and *then* throw to the caller, leading to redelivery and a second call that finds the activity already completed.

## Scope: analysis-first, code-change-if-needed

This issue is **investigation only** in its first phase. Expected output is a written audit comment on this issue:

1. **Map the call sites.** Every grain method invoked from a stream handler (custom task, script, condition evaluation, completion condition) that mutates `WorkflowInstance` state. Includes at minimum:
   - `IWorkflowInstanceCallback.CompleteActivity`
   - `IWorkflowInstanceCallback.FailActivity`
   - The narrower call surface from `ScriptExecutorGrain` and condition evaluators (those return values, but a redelivered evaluation could still call back into the workflow instance).
2. **For each, classify the redelivery behavior:**
   - **Naturally idempotent** — second call finds activity already in target state and no-ops. Document the guard.
   - **Idempotent by accident** — works today but only because of a specific guard that isn't obviously meant for this. Flag for hardening.
   - **Non-idempotent** — second call duplicates side effects (variable merge applied twice, compensation re-spawned, etc.). Must be fixed.
3. **Cross-reference with the event-sourcing model.** WorkflowInstance is a `JournaledGrain` — drained events go through `RaiseEvent → ConfirmEvents`. Each domain event needs an apply-twice analysis: if the same event is raised twice (e.g., `ActivityCompleted` for the same `ActivityInstanceId`), what's the projected state?
4. **The "Each activity instance executes at most once" invariant (CLAUDE.md design constraints)** is the relevant correctness target. The audit should confirm that the invariant survives stream redelivery.
5. **The compensation-walk variable-merge invariant** (also in CLAUDE.md design constraints) is the most likely place for a non-idempotency to hide — `VariablesMerged` raised twice would merge handler outputs into the parent scope twice. Special attention here.

## Phase 2 (conditional on Phase 1 findings)

If the audit finds non-idempotent paths:

- Add domain-level guards (compare-and-set on `ActivityInstanceId` state, or event deduplication by `(ActivityInstanceId, EventType)` in the journaled event log).
- Add a redelivery integration test using the Orleans `TestCluster`: force a stream redelivery by throwing in the handler after state mutation, assert end-state matches single-delivery outcome.

If the audit finds the system is genuinely idempotent:

- Replace the unsubstantiated comment at `CustomTaskHandlerBase.cs:122` with a precise reference to the guard that makes redelivery safe.
- Add a CLAUDE.md entry under design constraints documenting the idempotency contract so future contributors don't break it.

## Why now (priority)

Today's throughput is low enough that redelivery is rare. After #565 and #566 land, parallel workers process events on partitioned streams under real production load — any latent bug here becomes a production correctness incident, not a test-cluster flake.

This issue should land **before** #565/#566 are deployed to a load-test environment, even if Phase 2 produces no code changes.

## Out of scope

- Generalizing the audit to non-streaming RPC paths (Web → grain calls). Streaming is the dangerous path because Orleans retries silently; RPC errors surface to the caller.
- Building a generic event deduplication infrastructure. Start with targeted guards if needed.

## Acceptance

- Phase 1: audit comment posted to this issue covering every state-mutating call site reachable from a stream handler, with a verdict for each.
- Phase 2 (if needed): non-idempotent paths fixed with tests; or comment at `CustomTaskHandlerBase.cs:122` updated with a precise reference to the actual guard; CLAUDE.md design-constraints entry added.

## References

- `src/Fleans/Fleans.Worker/CustomTasks/CustomTaskHandlerBase.cs:121-124` (redelivery comment under audit)
- `src/Fleans/Fleans.Application/Grains/WorkflowInstance.cs` (state-mutation entry points)
- `src/Fleans/Fleans.Application/Grains/WorkflowInstance.Infrastructure.cs` (`DrainAndRaiseEvents` — event-sourcing apply path)
- CLAUDE.md design constraints: "Each activity instance executes at most once", "Compensation handlers run in isolated child scopes", "Registration-vs-cleanup error asymmetry (#425)"
- Related: #565, #566 (raise the stakes by raising throughput)

nightBaker:
## Summary

Closes #569 — Phase 2B (system is idempotent; docs-only fix).

Phase 1 audit ([verdict comment](https://github.com/nightBaker/fleans/issues/569#issuecomment-4468273751)) walked the 5 `[ImplicitStreamSubscription]` handlers (4 in `Fleans.Application/Events/Handlers/` + `CustomTaskHandlerBase`), enumerated 9 state-mutating call sites, and applied the v2 plan's 6-column completeness bar plus a compensation-walk deep-dive enumerating all 7 compensation event records at `WorkflowDomainEvents.cs:86-117`. Every row was `naturally-idempotent` — no `accidentally-idempotent`, no `non-idempotent`.

Phase 2B is the documentation lockdown:

- **`Fleans.Worker/CustomTasks/CustomTaskHandlerBase.cs:119`** — replaces the unsubstantiated `// let stream provider retry — domain idempotency guards handle duplicates` comment with a specific reference to the guards (`HasActiveEntry` at `WorkflowInstance.cs:315`/`:351`, defense-in-depth `entry.IsCompleted` at `WorkflowExecution.cs:2135`/`:2186`) and a pointer to the CLAUDE.md contract bullet.
- **`CLAUDE.md` Design Constraints** — adds a new "Stream-redelivery idempotency contract (#569)" bullet listing every guard the audit relied on (5 distinct guards for the 9 call sites + the projection-level idempotency of `ConditionSequenceEvaluated` and `VariablesMerged`), the compensation-walk single-entry protection, and the rule for future contributors: extending `IWorkflowInstanceCallback` (or any stream-reachable grain interface) requires extending the guard set or proving the new path is naturally idempotent.

No code change in production paths — the catch-and-rethrow at `CustomTaskHandlerBase.cs:121` is preserved verbatim.

## Test plan

- [x] `dotnet build src/Fleans/Fleans.Worker/Fleans.Worker.csproj` — clean (0 errors)
- [x] Comment-only + markdown-only changes; no test impact
- [ ] Manual regression: not required (docs-only); however reviewers can confirm the cited line numbers against the PR HEAD


nightBaker (commented):
## Code Review — docs(#569): pin stream-redelivery idempotency contract

🟢 Minor only. Verified all 8 cited file:line references against HEAD:

- `WorkflowInstance.cs:315/:351/:389` — `HasActiveEntry` guard + `joinState.HasFired` guard, all in place
- `WorkflowExecution.cs:1813/:2135/:2186` — `entry.IsCompleted` stale-guards on CompleteActivity / FailActivity / CompleteMultiInstanceEarly
- `WorkflowInstanceState.cs:393` — `SetConditionSequenceResult` (set-by-key)
- `WorkflowVariablesState.cs:31` — `Merge` (set-by-key)

Diff is comment-only on `CustomTaskHandlerBase.cs` (the `throw;` statement preserved verbatim) plus one new Design Constraints bullet in `CLAUDE.md`. Build passes; no test changes needed since there is no behavior change. The audit comment at #569 is the load-bearing artifact; this PR codifies the contract for future contributors.

### 🟢 Minor

1. **Line numbers in code comments rot fast.** `CustomTaskHandlerBase.cs:121-125` cites four file:line references (`WorkflowInstance.cs:315/:351`, `WorkflowExecution.cs:2135/:2186`). Any future refactor that shifts those lines makes the comment lie silently. The same risk exists in the new CLAUDE.md bullet — both rot together, which is fine, but consider shortening the *code* comment to a stable pointer (`// see CLAUDE.md Design Constraints "Stream-redelivery idempotency contract"`) and keeping the exhaustive file:line list in CLAUDE.md where any grep/refactor that touches the guards will land. One canonical place to update beats two.

2. **CLAUDE.md bullet density is consistent with the neighboring `#425` bullet** (Registration-vs-cleanup error asymmetry, directly above), so the format is pattern-matched to project precedent. No action needed; just calling out that the precedent is intentional.

3. **The audit comment is the durable artifact** — verdict 2B template, 9-row table, compensation deep-dive table. PR description correctly links to it. Reviewers who want full per-call-site analysis should follow the link rather than re-deriving from the diff.

Issue #591, opened 2026-05-16
Stream sharding: activation budget + PubSubStore growth/cleanup (#565 follow-up)


nightBaker:
## Problem

Companion gap for #565. Sharding the stream Guid by `WorkflowInstanceId` activates one implicit-subscription grain **per (handler type × workflow instance)**. The current bottleneck has worst case ≈ 1 activation per event type; after #565 the worst case is `O(active_instances × handler_types)`.

Concrete: 1k concurrent instances × 5 plugin types (custom task) + 4 fixed event types (script, condition, activation condition, completion condition, IDomainEvent) ≈ 9k grain activations vs. ≈ 9 today. Each implicit-subscription handle also persists state in Redis `PubSubStore` (clustering & pub-sub use the same Redis we provision). Without an explicit cleanup story, `PubSubStore` keys grow monotonically with every workflow instance ever created.

#565 acceptance does not budget memory pressure, activation-collection tuning, or `PubSubStore` GC.

## Proposed work

1. Measure activation count + Redis key count under a representative load (use the load-test profile referenced by the May 16 azure load-test PR).
2. Tune Orleans activation collection (`CollectionQuantum`, `CollectionAge`) and `[KeepAlive]` policy so idle plugin-grain activations release within a bounded window after workflow completion.
3. Add `PubSubStore` cleanup on `WorkflowInstanceCompleted` / `WorkflowInstanceFailed` — explicitly drop subscriptions tied to the terminated instance Guid so Redis doesn't grow unbounded.
4. Add a metric (event id range 7000+) for active implicit-subscription grain count and `PubSubStore` key count, surfaced on the Orleans dashboard.

## Acceptance

- Load test: 1k instances × 5 plugin types complete with bounded memory and bounded Redis `PubSubStore` key count (key count returns to baseline within N minutes after workflows complete).
- Cleanup is exercised in `WorkflowInstance` deactivation path; confirmed by an integration test.

## Refs

- #565 — depends on its sharding strategy
- src/Fleans/Fleans.ServiceDefaults/FleanStreamingExtensions.cs:53 (`TotalQueueCount=8`)
- src/Fleans/Fleans.Application/Grains/WorkflowInstance.cs (deactivation path)

nightBaker:
## Review of design

[...]

PR #617, opened 2026-05-19
feat(#591): Phase A0 — stream-sharding observability metrics

nightBaker:
## Summary

First PR for #591 (P1). Implements **Phase A0** of the approved phased design: a hosted service that polls every 30s and emits two structured log lines giving operators visibility into the per-instance stream-sharding cost from #565.

**Phase A1** (TestCluster `PubSubStore` lifecycle probe), **A2** (load-test harness in `tests/load/stream-sharding-pressure/`), and **A3** (numeric pass criterion to gate Phase B/C) are deferred to follow-up PRs within #591. A0 alone is independently valuable — operators can observe the activation explosion and PubSubStore growth before any tuning work commits.

## What changed

| File | Change |
|---|---|
| `src/Fleans/Fleans.ServiceDefaults/Observability/ImplicitSubMetricsHostedService.cs` | New `BackgroundService` (110 lines) polling Orleans `IManagementGrain.GetSimpleGrainStatistics()` + Redis `SCAN MATCH "PubSubStore/*"`. Emits 3 `[LoggerMessage]` declarations: EventId 7000 / 7001 / 7099. |
| `src/Fleans/Fleans.ServiceDefaults/Observability/ImplicitSubMetricsExtensions.cs` | `AddImplicitSubMetrics()` extension method registering the hosted service as singleton. |
| `src/Fleans/Fleans.Api/Program.cs` | Calls `AddImplicitSubMetrics()` after `AddApplication()` / `AddInfrastructure()`. |
| `src/Fleans/Fleans.WorkerHost/Program.cs` | Same. |
| `CLAUDE.md` | New bullet in *Things to Know* documenting EventId 7000 / 7001 / 7099 + scope-context. |

## EventId summary

| EventId | Level | Message |
|---|---|---|
| 7000 | Information | `Implicit-subscription handler-grain activation count: {TotalActivations} (breakdown: {PerTypeBreakdown})` — sums grain counts filtered to types whose name contains `Handler` (`*EventHandler`, `CustomTaskHandler*`). Per-type breakdown lists only types with nonzero activations. |
| 7001 | Information | `Orleans PubSubStore Redis key count: {KeyCount}` — emitted only when `IConnectionMultiplexer` is in DI (production: yes via Aspire/Redis; Memory-store tests: skipped). |
| 7099 | Warning | `ImplicitSubMetricsHostedService poll iteration failed; will retry on next interval` — per-iteration failures don't stop the host. |

## Design decisions

- **30s polling interval** as a starting point. Configurable knob deferred to A2/A3 — no need until load-testing shows the cadence is wrong.
- **No `IGrainStorage` introspection for Memory PubSubStore.** Crossing the `IGrainStorage` abstraction to count Memory-store entries was rejected as test scaffolding leaking into production. Production runs against Redis (verified: `Fleans.Api/Program.cs:84`, `Fleans.WorkerHost/Program.cs:53`); Memory-store paths are dev/test-only and don't need a key-count metric.
- **5s startup delay** before the first poll. Orleans `IManagementGrain` is not reliably callable during the first second of silo lifetime.

## Test plan

- [x] `dotnet build Fleans.sln` — 0 errors (587 pre-existing warnings, unchanged shape).
- [x] CI build + pg-tests on this PR.
- [ ] Manual smoke (out-of-band, single command): start Aspire stack, run any workflow, grep API/Worker logs for `EventId 7000` and `EventId 7001`. Both should appear at 30s cadence.

## Refs

- Round-2 approved [design](https://github.com/nightBaker/fleans/issues/591#issuecomment-4484492129) — this PR implements step A0.
- Round-2 [approval](https://github.com/nightBaker/fleans/issues/591#issuecomment-4484503151).
- Depends on #565 stream-sharding (already in `main`).

Refs #591 (does not close — Phase A1/A2/A3 remain).


nightBaker (commented):
## Code Review — feat(#591): Phase A0 — stream-sharding observability metrics

Scope: 148 insertions across 5 files (2 new files in `Fleans.ServiceDefaults/Observability/`, 1-line CLAUDE.md bullet, 2× `AddImplicitSubMetrics()` registration in `Fleans.Api`/`Fleans.WorkerHost` `Program.cs`). Self-contained additive change; nothing existing modified beyond the registration lines.

### Verified

| Check | Status | Evidence |
|---|---|---|
| Orleans 10.x API correctness | ✅ | `_clusterClient.GetGrain<IManagementGrain>(0).GetSimpleGrainStatistics()` is the current management-grain pattern. |
| `[LoggerMessage]` source-generator usage | ✅ | All three log methods declared `private partial void` on the partial class — matches the project-wide CLAUDE.md rule. |
| EventId allocation (7000/7001/7099) | ✅ | New range, no collision with the existing `WorkflowInstance` log EventIds (1xxx–4xxx) or `WorkflowEventsPublisher` 5001. Documented in CLAUDE.md edit. |
| Redis SCAN pattern (not KEYS) | ✅ | `server.KeysAsync(pattern: "PubSubStore/*", pageSize: 1000)` translates to SCAN under StackExchange.Redis — won't block the server on large keyspaces. |
| Cancellation discipline | ✅ | Both `ExecuteAsync` `Task.Delay` calls wrapped in `try/catch (OperationCanceledException) { return; }`; `KeysAsync` chained with `WithCancellation(ct)`. Clean shutdown. |
| Idempotent registration | ✅ | `services.AddHostedService<T>()` deduplicates by type, so the two `Program.cs` registrations on `Fleans.Api` + `Fleans.WorkerHost` (separate processes) don't double-emit. |
| Scope honesty for Phase A0 | ✅ | PR description explicitly defers A1/A2/A3 to follow-up PRs within #591. No-test acceptable because the metric is operationally validated, not logically invariant. |
| Optional `IConnectionMultiplexer` resolution path | ✅ | `Fleans.ServiceDefaults/FleanStreamingExtensions.cs:50` bridges `AddKeyedRedisClient("orleans-redis")` → non-keyed singleton when `Fleans:Streaming:Provider=redis` (the default). Constructor's `IConnectionMultiplexer? redis = null` resolves correctly via DI in the default deployment. |

### 🔴 Blocking
None.

### 🟡 Important
1. **EventId 7001 only fires on the default Redis streaming path.** The non-keyed `IConnectionMultiplexer` bridge in `FleanStreamingExtensions.AddRedisStreams` is only registered when `Fleans:Streaming:Provider=redis`. Deployments running `FLEANS_STREAMING_PROVIDER=Kafka` or `=AzureQueue` (still on Redis-backed `PubSubStore`, since `PubSubStore` is always Redis) will see `_redis == null` and silently skip the key-count probe — losing half the observability surface that #591 Phase A is supposed to deliver. The constructor comment ("Memory-store test paths, Aspire dev with Memory streams") covers Memory but not Kafka/AzureQueue. A future-proof alternative would be `[FromKeyedServices("orleans-redis")] IConnectionMultiplexer? redis = null` — the keyed registration is always present whenever Redis is the `PubSubStore` backend. **Not blocking** because (a) Redis streaming is the documented default, (b) Phase A1's empirical probe test will catch this gap, and (c) the PR scope is explicitly Phase A0. Recommend filing as a Phase A0 follow-up note on the #591 issue so it's not lost.

### 🟢 Minor
1. **Filter chain redundancy** (`ImplicitSubMetricsHostedService.cs:69-71`): `s.GrainType.Contains("EventHandler") || s.GrainType.Contains("CustomTaskHandler") || s.GrainType.Contains("Handler")` — the third predicate strictly dominates the first two ("EventHandler" and "CustomTaskHandler" both contain "Handler"). Collapsing to a single `Contains("Handler")` is equivalent. Keeping the explicit list as documentation-via-code is also defensible; either way is fine.
2. **Heuristic breadth**: the `Contains("Handler")` filter will also include any non-implicit-subscription grain whose type name happens to contain `Handler` (e.g., a future `HandlerCacheGrain`). Acceptable for now — the per-type breakdown in EventId 7000 lets operators visually distinguish them.
3. **Single-server Redis assumption**: `_redis.GetServers().FirstOrDefault(s => s.IsConnected)` counts keys on one server only. Correct for Fleans's documented topology (single Redis instance for clustering + `PubSubStore`). If the project ever ships Redis Cluster guidance, this becomes an undercount and the iteration would need to span all servers.
4. **5-second startup delay** before first probe is well-commented; consider re-validating once the project standardises on a silo readiness signal (out of scope here).

No blocking issues. OK to promote (gates on CI green + Step 8 marker-freshness check on the next worker iteration).

PR #619, opened 2026-05-19
feat(#588): user-task claiming by user group

### LINKED ISSUE #588: usertask claiming by user group

nightBaker:
## Summary

Closes the silent-authorization-bypass on group-only user tasks: a `<bpmn:userTask>` gated by `camunda:candidateGroups` (no assignee, no candidateUsers) was claimable by **any authenticated caller** because both pre-existing checks (`matchesAssignee`, `matchesCandidateUsers`) passed trivially when their respective constraints were absent. This was a real authz hole, not a missing feature.

### Domain

`UserTaskLifecycle.Claim` rewritten as a flat three-rule "OR": assignee match OR candidateUsers contains OR `CandidateGroups ∩ userGroups ≠ ∅` (ordinal case-sensitive). Empty constraints still mean "unrestricted". Rejection messages consolidated to `"User {userId} is not authorized to claim this task"` — identifier-free, so the task's authorization shape doesn't leak via exception bodies (the prior message leaked the `Assignee` identifier).

### Caller-side groups source

New `IUserGroupResolver` service (`Fleans.Api/Authorization/`) with two implementations, registered by `Program.cs` based on `Authentication:Authority`. Mirrors the `IUserTaskFilterStrategy` two-impl-by-config precedent.

- `BodyUserGroupResolver` — reads `ClaimTaskRequest.UserGroups` verbatim when auth is disabled (default Fleans posture).
- `JwtUserGroupResolver` — reads `HttpContext.User.FindAll("groups")` (OIDC repeated-claim convention; claim name overridable via `Authentication:GroupsClaim`) when auth is enabled. Body-supplied `UserGroups` is **ignored** so callers can't spoof groups they don't have.

JSON-array-encoded JWT groups are NOT auto-parsed — operators with array-encoded IdPs need to either configure the IdP to repeat the claim or wait for a follow-up issue tracking JSON-array support (filed separately).

### DTO

`ClaimTaskRequest` extended to `(string UserId, IReadOnlyList<string>? UserGroups = null)`. Default-null preserves wire-format backward compatibility — existing callers that POST `{"UserId":"alice"}` keep working for tasks without `CandidateGroups`.

### Audit logging

EventId 1066 `LogUserTaskClaimRejected` (Warning) fires per rejected claim, carrying `UserGroupCount` only — no group names — so rejection counts are diagnostically visible without leaking sensitive identifiers. Allocated the next free slot in the user-task lifecycle range 1060-1069.

### Tests

- 6 new domain unit tests in `UserTaskLifecycleTests`: only-candidateGroups match/no-match/empty-userGroups, assignee+groups (each rule wins independently), ordinal case-sensitivity guard.
- Existing `UserTaskLifecycleTests` / `UserTaskIntegrationTests` callers updated to pass `[]` for the new `userGroups` parameter (mechanical).
- **51 user-task tests pass** (28 domain + 23 integration).

### Manual test plan #61

`tests/manual/61-usertask-group-claim/test-plan.md` registered with a `camunda:candidateGroups`-only BPMN fixture. 5 verification steps: happy-path claim, rejection-path 409, audit-log line check, backward-compat for unrestricted tasks.

### CLAUDE.md

New "Design Constraints" entry documenting the three-rule "OR" model, the resolver split, and the audit-log shape — so future contributors don't re-invent the AND-vs-OR confusion that produced the original bug.

## Test plan

- [x] `dotnet build Fleans.sln` — 0 errors.
- [x] `dotnet test Fleans.Domain.Tests --filter UserTask` — 28/28 pass.
- [x] `dotnet test Fleans.Application.Tests --filter UserTask` — 23/23 pass.
- [ ] Manual test plan #61 (5 steps) — to be exercised in regression.

Closes #588.


nightBaker (commented):
## Code Review — feat(#588): user-task claiming by user group

Scope: 391 insertions / 55 deletions across 19 files. The domain change is the security-critical core (47 lines on `UserTaskLifecycle.cs`); everything else is plumbing, tests, plan, and docs.

### Verified

| Check | Status | Evidence |
|---|---|---|
| Silent-bypass closure | ✅ | Before: a task with only `CandidateGroups` set passed both pre-existing guards trivially (`metadata.Assignee is null` → `matchesAssignee = true`, etc.) → any caller claimable. After: flat three-rule "OR" — `anyConstraint && !(any-match)` throws. `CandidateGroups`-only path now hits the new `matchesCandidateGroups` branch. |
| Three-rule "OR" semantics | ✅ | `UserTaskLifecycle.cs:22-49` — `hasAssignee` / `hasCandidateUsers` / `hasCandidateGroups` flags + `anyConstraint` guard + final `if (anyConstraint && !(matchesAssignee || matchesCandidateUsers || matchesCandidateGroups))` throw. Empty-constraints path (no rule present) falls through to allow, preserving "unrestricted task" semantics for backward compat. |
| Identifier-free rejection message | ✅ | Consolidated to `"User {userId} is not authorized to claim this task"`. Prior assignee-leaking variants (`"Task is assigned to {assignee}, not {userId}"`) gone. |
| `IUserGroupResolver` two-impl pattern | ✅ | Mirrors `IUserTaskFilterStrategy` precedent (`Fleans.Persistence/IUserTaskFilterStrategy.cs`). Registered by config in `Fleans.Api/Program.cs:50-56` based on `Authentication:Authority`. |
| Body-spoof prevention in auth-enabled path | ✅ | `JwtUserGroupResolver.Resolve` reads only from `HttpContext.User.FindAll(_claimName)`. `ClaimTaskRequest.UserGroups` is never consulted on that path — authenticated callers cannot inject groups via the body. |
| Unauthenticated defensive guard in JWT resolver | ✅ | `if (httpContext.User?.Identity?.IsAuthenticated != true) return Array.Empty<string>()` — drives the domain-layer rejection rather than throwing, so the response surface stays consistent. |
| OIDC repeated-claim convention | ✅ | `FindAll(_claimName).Select(c => c.Value)` — one entry per repeated claim. JSON-array variant explicitly deferred per design open-question 3. |
| Configurable claim name | ✅ | `_claimName = configuration["Authentication:GroupsClaim"] ?? "groups"` — operator can pin to `roles` for Entra ID app-roles deployments. |
| `ClaimTaskRequest` wire-format compat | ✅ | `record ClaimTaskRequest(string UserId, IReadOnlyList<string>? UserGroups = null)` — `null`-defaulted optional preserves existing `{"UserId":"alice"}` callers. |
| Audit-log shape (EventId 1066) | ✅ | `LogUserTaskClaimRejected(activityInstanceId, userId, userGroupCount)` — counts only, no group-name leak. Warning level, properly named. EventId in the documented user-task range 1060-1069 (next free slot). |
| Grain catch-and-rethrow preserves InvalidOperationException → controller's 409 mapping | ✅ | `WorkflowInstance.ClaimUserTask` re-throws after logging; controller catches `InvalidOperationException` and returns `Conflict(ex.Message)` (controller path unchanged at this level). |
| BPMN fixture matches parser | ✅ | `tests/manual/61-usertask-group-claim/group-claim.bpmn` uses `camunda:candidateGroups="managers,supervisors"` as a direct attribute on `<bpmn:userTask>` with `xmlns:camunda` declaration — matches `BpmnConverter.cs:314-315` and #18 precedent. |
| Test coverage | ✅ | 6 new domain tests + 23 existing integration tests pass. Filter `UserTask` slice: 28/28 domain + 23/23 integration (51 total). |
| Existing callers cleanly migrated | ✅ | 18 ClaimUserTask call sites in tests updated to `(…, …, [])` (mechanical sed). No production callers besides the grain interface chain — verified by `grep -rn 'ClaimUserTask'` returning only domain/grain/test paths. |
| CLAUDE.md anchor | ✅ | New "Design Constraints" bullet describes the OR rule + resolver split + audit-log shape. Future contributors reading the file before touching the claim path will see the invariant. |

### Security-posture check (independent re-read)

The pre-existing code had an actual silent authz bypass for the canonical Camunda/Zeebe authoring shape (`<bpmn:userTask camunda:candidateGroups="…">` with no assignee/candidateUsers). Confirmed by re-reading the old code: when `metadata.Assignee is null` AND `metadata.CandidateUsers.Count == 0`, both `matchesAssignee` and `matchesCandidateUsers` evaluated `true` trivially, and the "only one constraint set" branch never fired because the actual `CandidateGroups` constraint wasn't even considered. **The new code closes that path correctly.** I verified by mental execution against the new logic with `CandidateGroups=["managers"]`, `userGroups=[]`: `hasAssignee=false`, `hasCandidateUsers=false`, `hasCandidateGroups=true` → `anyConstraint=true`; `matchesCandidateGroups=false` → throws. ✓

### Scope honesty

The PR description correctly flags the JSON-array JWT-claim deferral as a follow-up. No work claimed beyond what's in the diff.

### 🔴 Blocking
None.

### 🟡 Important
None.

### 🟢 Minor — non-blocking

1. **`JwtUserGroupResolver` does not validate that the controller's authz pipeline ran**. The defensive `IsAuthenticated != true → empty list` guard handles the case, but a future change to the controller (or a middleware reordering) could land unauthenticated requests with `httpContext.User == null`. The current null-conditional `httpContext.User?.Identity?.IsAuthenticated` handles that too, but a follow-up could add `ArgumentNullException.ThrowIfNull(httpContext)` for stricter contract. Trivial.

2. **`Array.Empty<string>()` vs `[]`**: the new code uses both forms inconsistently (`BodyUserGroupResolver` uses `Array.Empty<string>()`, tests use `[]`). Both compile to the same thing under collection-expression target typing for `IReadOnlyList<string>`. Stylistic — pick one in a future cleanup pass.

3. **Manual plan #61 step 5 references `$TASK_C`** but doesn't define it (assumes the operator deploys a separate unrestricted fixture). Minor — plan reads as "exercise these scenarios" not a copy-pasteable script. Could be clarified with a callout: "Step 5 requires deploying an additional fixture without `CandidateGroups` — use the existing #18 fixture as your `$TASK_C` source."

4. **Round-2-review minor #1 deferred-test-scope decision**: the round-2 analysis-review recommended `IUserGroupResolver` unit tests (one for each resolver). This PR ships the two impls but does NOT include unit tests for them. Recommendation: leave this PR as-is (the resolvers are 17/40 lines each, the dispatch logic is trivial, and the integration test path through `Fleans.Application.Tests` exercises the body-path branch). The JWT-path unit test is the only meaningful gap; a follow-up issue can capture it if a reviewer wants stronger coverage. Filing the follow-up at merge time would close the loop cleanly.

No blocking issues. OK to promote (gates on the Step-8 fresh-marker check on the next worker iteration).

PR #620, opened 2026-05-19
chart(streaming): wire Fleans__Streaming__Provider for all four providers (#599)

nightBaker:
## Summary

Closes the silent-fallback footgun in `charts/fleans/templates/_helpers.tpl`: pre-fix, only `streaming.provider=Kafka` emitted `Fleans__Streaming__Provider` at all — operators setting `Redis`, `AzureQueue`, or `Memory` saw no env var rendered, and the silo silently ran whatever the engine default was (Redis). Worse, the chart's own default was `Memory` while the engine's was `Redis`, so the no-set deployment shape produced the same wrong outcome.

The new switch covers all four supported providers explicitly with case-insensitive matching on the input and canonical capitalised values on the emitted env var. An unknown value (typo, or a future provider not yet wired into the chart) aborts `helm template` at render time via `{{- fail }}` — typos can't ship to a cluster as broken pods.

### Behavior change: chart default Memory → Redis

The values.yaml default is rebased to match the engine's `?? "redis"` default. Memory is single-silo debug-only per CLAUDE.md *Four stream providers*; shipping it as the chart default silently downgraded any multi-silo deployment. Operators who relied on the prior chart default (which was always broken in multi-silo shapes) will silently flip to Redis on upgrade — flagged in the new "Streaming provider" section of `charts/fleans/README.md`.

### Files

- `charts/fleans/templates/_helpers.tpl` — 4-branch `if/else if` switch + `{{- fail }}` validation guard.
- `charts/fleans/values.yaml` — new `streaming.azureQueue.connectionString` key; default `provider: Redis`; Memory warning comment above the key.
- `charts/fleans/README.md` — new "Streaming provider" subsection with provider table and Memory warning at the top.
- `tests/manual/62-chart-streaming-providers/test-plan.md` — 7 `helm template | grep` assertions.
- `tests/manual/README.md` — catalog entry 62.

## Test plan

All 7 manual steps in plan #62 verified locally against this branch with helm v4.1.4:

- [x] Step 1: `--set streaming.provider=Redis` → emits `Fleans__Streaming__Provider="Redis"` on every workload, no `__Kafka__` or `__AzureQueue__` env vars.
- [x] Step 2: `--set streaming.provider=Kafka --set streaming.kafka.brokers=...` → emits both env vars (Kafka regression check).
- [x] Step 3: `--set streaming.provider=AzureQueue --set streaming.azureQueue.connectionString=...` → emits both env vars.
- [x] Step 4: no `--set` → chart default emits `Fleans__Streaming__Provider="Redis"` (the rebased default).
- [x] Step 5: `--set streaming.provider=redis` (lowercase) → emits `"Redis"` (canonical capitalised form, even though input was lowercase).
- [x] Step 6: `--set streaming.provider=Memory` → emits `Fleans__Streaming__Provider="Memory"` (pre-fix this was nothing — the round-1 silent-fallback hole).
- [x] Step 7: `--set streaming.provider=Reddis` → `helm template` exits non-zero with `"Unsupported streaming.provider \"Reddis\". Valid: Memory, Redis, Kafka, AzureQueue (case-insensitive)."`

## Follow-up tracked separately

- AzureQueue `connectionString` `existingSecret` pattern (matches the `auth.clientSecret` precedent — currently inline, should be Kubernetes Secret-sourced for production).

Closes #599.


nightBaker (commented):
## Code Review — chart(streaming): wire Fleans__Streaming__Provider for all four providers (#599)

Scope: 155 insertions across 5 files. Chart-only change; no application/domain/Orleans surface touched. The implementation directly matches the round-2 approved plan.

### Verified

| Check | Status | Evidence |
|---|---|---|
| All 4 providers emit explicit env var | ✅ | Re-ran `helm template ... --set streaming.provider=X` for X in {Memory, Redis, Kafka, AzureQueue} on the PR branch. Every case produced `Fleans__Streaming__Provider="<canonical>"`. |
| Case-insensitive on input, canonical on emit | ✅ | `--set streaming.provider=redis` → emits `"Redis"`. Match via `lower $provider`, emit is the literal canonical string. |
| `{{- fail }}` validation guard fires on typos | ✅ | `--set streaming.provider=Reddis` → `helm template` exits non-zero with `Unsupported streaming.provider "Reddis". Valid: Memory, Redis, Kafka, AzureQueue (case-insensitive).` Renders before any pod schedules. |
| Chart default rebased Memory → Redis | ✅ | `helm template fleans charts/fleans` (no `--set`) emits `Fleans__Streaming__Provider="Redis"`. values.yaml line 209: `provider: Redis`. |
| Kafka regression intact | ✅ | `--set streaming.provider=Kafka --set streaming.kafka.brokers=...` emits both `Fleans__Streaming__Provider="Kafka"` AND `Fleans__Streaming__Kafka__Brokers="..."`. |
| AzureQueue connection string env var | ✅ | `--set streaming.provider=AzureQueue --set streaming.azureQueue.connectionString=...` emits both env vars. Name matches `Fleans.Aspire/Program.cs:154` precedent. |
| All workloads consume the helper | ✅ | `helm template ... --set worker.enabled=true` shows `Fleans__Streaming__Provider` rendered 4 times (api + web + mcp + worker), matching the 4 deployment templates consuming `fleans.commonEnv`. |
| `helm lint` clean | ✅ | `helm lint charts/fleans` → `0 chart(s) failed`. |
| Memory single-silo warning visible | ✅ | values.yaml comment above `provider: Redis`: `"WARNING: Memory is single-silo debug-only — DO NOT use in multi-silo deployments."` Repeated as a callout in chart README's new "Streaming provider" subsection. |
| Manual plan #62 with 7 assertion steps | ✅ | Plan covers happy paths, default, lowercase, Memory-explicit-emit (the round-1 regression target), and the typo path. README catalog entry registered between #61 and Website suite. |
| README streaming-provider section placement | ✅ | New `### Streaming provider` subsection lands between OIDC and Worker-silos guidance, matching the chart's logical config-section ordering. Memory warning is in a top-level callout (per round-1 🟢 #6 fix). |
| No breakage to existing chart paths | ✅ | `_helpers.tpl` changes are bounded to the streaming block (lines 109-113 of pre-fix code → 109-128 post-fix). Persistence + Redis-URL paths above and below untouched. |

### Engine-side consistency check

The values emitted (`"Memory"`, `"Redis"`, `"Kafka"`, `"AzureQueue"`) match the engine's `Fleans.ServiceDefaults/FleanStreamingExtensions.cs:30-39` switch arms exactly (after engine-side `ToLowerInvariant()`). No engine-side parser change needed.

### 🔴 Blocking
None.

### 🟡 Important
None.

### 🟢 Minor

1. **`list` duplication between the `has` guard and the canonical-form switch.** Round-2 review #2 flagged this; the design's YAGNI defense is reasonable for four entries. A future maintainer adding a fifth provider must update both sites. Optional follow-up to hoist `{{- $supported := list "memory" ... }}` to the top of the helper.

2. **`fail` is the chart's first Sprig `fail` usage** — no precedent to cross-reference. The implementation is standard Sprig (`printf` for message interpolation, no logic). Documenting one usage of a Sprig builtin doesn't merit a CLAUDE.md bullet, but a reviewer of the next chart-rule addition might want to know this exists.

3. **The values.yaml AzureQueue `connectionString` is inline** with a comment pointing at the follow-up. Matches the `auth.clientSecret` precedent. Operators wanting Kubernetes-native secret sourcing must use `extraEnv` until the follow-up lands.

4. **Memory single-silo warning lives in three places now** — values.yaml comment, chart README, and the `fail` message lists Memory as a valid value but the surrounding docs warn against it. Slight friction: an operator who reads only the `fail` message sees "Memory is valid" without the "but don't use it in multi-silo" caveat. Acceptable — the validation guard is for typos, not for nudging operators away from a documented-debug-only choice. Flagging only for completeness.

No blocking issues. OK to promote (gates on the Step-8 fresh-marker check + CI on the next worker iteration).

PR #635, opened 2026-05-20
ci(#634): Phase 0 — website-pin-guard CI check + allow-list ratchet

nightBaker:
## Summary

Phase 0 of #634 (CLAUDE.md *Documentation scope split* cleanup). Lands the CI guard + ratchet allow-list **before** any prose-rewrite phase so the cleanup PRs can't be undone by a parallel landing.

### `.github/workflows/website-pin-guard.yml`

Runs on every PR touching `website/src/content/docs/**`, `tools/website-pin-allowlist.txt`, or the workflow itself. Fails on:
- **NEW violations** — `.cs:NNN` line pins or inline `src/Fleans/...` paths that aren't in the allow-list (regression check).
- **STALE allow-list entries** — lines listed in the allow-list but no longer present in the docs (forces the file to shrink as phases land).

### `tools/website-pin-allowlist.txt`

101 known violations enumerated as `file:line:content` tuples. Self-destructs in Phase 3 (per #634 round-2 plan) — each Phase 1/2 PR removes the entries it eliminates.

## Carve-outs (NOT counted as violations)

- **`github.com/<owner>/<repo>/blob/<ref>/...src/Fleans/...` URLs** — clickable source links that don't require a checkout. Verified user-facing in `self-host-helm.md`, `self-host-docker-compose.md`, `multi-instance-activities.md`.
- **Type/method-name references** (e.g. `CustomTaskHandlerBase`, `AddCustomTaskPlugin`) — plugin authors consume these from the NuGet surface. The regex only matches `.cs:NNN` and `src/Fleans/...` paths, so type names pass naturally.

## Verification (local, before push)

- ✅ Allow-list matches current state exactly: 101/101 entries.
- ✅ Regex catches `.cs:NNN` line pins (`BpmnConverter.cs:9999` → match).
- ✅ Regex catches inline `src/Fleans/` (`src/Fleans/foo` → match).
- ✅ Regex correctly EXCLUDES `github.com/.../blob/...src/Fleans/` URLs (no match).

## Test plan

- [x] `grep -rEn '\.cs:[0-9]+|(^|[^/])src/Fleans/' website/src/content/docs/ | grep -vE 'github\.com/[^/]+/[^/]+/(blob|tree)/[^)]*src/Fleans/' | sort` — output matches `tools/website-pin-allowlist.txt` exactly.
- [ ] CI run on this PR — should pass (allow-list matches HEAD).
- [ ] Synthetic regression test: cherry-pick a sample new pin into a guide on a scratch commit, push, observe CI red — then revert. (Not in this PR; verified locally by injecting a fake violation into the diff and confirming the regex catches it.)

## What Phase 0 does NOT do

- Does **not** edit any website docs prose. The 101 violations stay in place; the guard just prevents the count from growing.
- Does **not** migrate `bpmn-support.md` to `docs/bpmn-coverage-matrix.md` (Phase 1).
- Does **not** strip pins or DRIFT-GUARD blocks from guides (Phase 2a/2b/2c).
- Does **not** harden the guard (Phase 3 — deletes the allow-list).
- Does **not** add the `reference/api.md` route refresh (Phase 2d — carved out per the round-2 review's 🟡 finding).

Refs #634


nightBaker (commented):
## Code Review — ci(#634): Phase 0 — website-pin-guard CI check + allow-list ratchet

Scope: 2 new files, 206 insertions, 0 deletions.

- `.github/workflows/website-pin-guard.yml` (105 lines) — runs on PRs touching website docs, the allow-list, or the workflow itself.
- `tools/website-pin-allowlist.txt` (101 lines) — frozen snapshot of known violations.

### Verified

| Check | Status | Evidence |
|---|---|---|
| Pin-guard CI check passes on this PR | ✅ | `gh pr checks 635` → `pin-guard pass 5s`. (First run failed on a trailing-newline diff artifact, fix shipped as `0d76ba7`.) |
| Allow-list count matches current scope | ✅ | 101 entries, matches the empirical grep against current `main` baseline that the round-2 analytic established. |
| Regex carve-out — `github.com/.../blob/...src/Fleans/` URLs preserved | ✅ | `grep -E '(^|[^/])src/Fleans/'` correctly excludes `/blob/main/src/Fleans/...` URLs because `/` precedes `src/`. Verified via `grep -rE 'github\.com/nightBaker/fleans/blob/[^)]*src/Fleans/' website/src/content/docs/` returning 3 user-facing URLs — none of them are in the allow-list (correctly carved out). |
| Regex catches new `.cs:NNN` line pins | ✅ | Echo-test: `echo "BpmnConverter.cs:9999" | grep -E '\.cs:[0-9]+'` matches. |
| Regex catches new inline `src/Fleans/...` paths | ✅ | Echo-test: `echo "src/Fleans/foo" | grep -E '(^|[^/])src/Fleans/'` matches. |
| Allow-list ratchet semantics | ✅ | Workflow's "Compare against allow-list" step fails on BOTH new violations (regression) AND stale allow-list entries (forces shrinkage). |
| Allow-list-deleted hardening path | ✅ | `if [ ! -f "$allowlist" ]` branch: any violation fails. Will activate when Phase 3 deletes the file. |
| Read-only permissions | ✅ | `permissions: contents: read` declared at the workflow level. |
| Triggers only on PR (not push to main) | ✅ | `on: pull_request` with `paths:` filter limiting to website docs + allow-list + workflow. |
| GitHub annotations (`::error::`) for failure output | ✅ | Each failure mode emits structured error annotations so PR UI surfaces them inline. |

### Phase 0 scope honesty

PR description correctly enumerates what's NOT done:
- No website docs prose touched (the 101 violations stay).
- No `bpmn-support.md` migration (Phase 1).
- No DRIFT-GUARD block removal (Phase 2).
- No guard hardening (Phase 3).
- No `reference/api.md` route refresh (Phase 2d, carved out per round-2 review's 🟡 finding).

### 🔴 Blocking
None.

### 🟡 Important
None.

### 🟢 Minor

1. **The `paths:` filter on `pull_request` triggers only on changes touching specific paths.** If a PR modifies *only* a file in `src/Fleans/` and that change accidentally renames something the website cites by path (e.g. renames `BpmnConverter.cs`), the website pin would now point at a non-existent file but the guard wouldn't run because the PR doesn't touch website docs. Acceptable for Phase 0 — that's a stale-pin-after-source-rename concern (the broader "drift-guard" semantics), not a "new violation introduced" concern. The guard is intentionally scoped to "did someone ADD source-code references to website docs"; the inverse problem (source moves out from under existing pins) is what the DRIFT-GUARD HTML blocks were trying to catch and will be handled by Phase 1's migration to `docs/bpmn-coverage-matrix.md`.

2. **The workflow uses `gh`-style `::error::` annotations** which work in any GitHub-hosted runner — well chosen. No `action/cache` or other secondary actions, no third-party action dependencies — Node version churn (`actions/checkout@v4` is the only dependency; flagged as Node 20 deprecation in other workflows but `@v4` is stable through 2026 per the GitHub roadmap). Low maintenance footprint.

3. **`continue-on-error` vs hard fail**: the workflow uses hard `exit 1` on any mismatch. That's the right choice — Phase 0's whole point is to PREVENT regressions. If a future phase needs to land a non-pin-related change to a guide that incidentally shifts line numbers in the allow-list, the developer has to update the allow-list as part of that PR. Friction, but intentional.

4. **`tools/` is a new top-level directory** introduced by this PR. No existing `tools/` infrastructure — first user. Acceptable; `tools/` is a conventional name for repo-local build/dev tooling and the dir's purpose is self-evident from its single file.

No blocking issues. OK to promote (gates on Step-8 fresh-marker check + the .NET build CI completing on the next worker iteration).

PR #648, opened 2026-05-20
feat(#457): PlacementRoleAssertion fails fast on Fleans:Role / placement-attribute mismatch

nightBaker:
## Summary

Closes #457 — adds `PlacementRoleAssertion` to fail fast at silo startup when `Fleans:Role` is incompatible with a registered grain's `[WorkerPlacement]` / `[CorePlacement]` attribute.

3 rounds of analytic ↔ analysis-review converged on the Option A architecture (per-app extension call from `Fleans.Worker.Hosting`, layering-clean). This PR implements that design.

## How it works

A `public sealed PlacementRoleAssertion : ILifecycleParticipant<ISiloLifecycle>` registers at `ServiceLifecycleStage.RuntimeInitialize` and:

1. Reads `Fleans:Role` from `IConfiguration` and the running `ILocalSiloDetails.Name`.
2. Scans `AppDomain.CurrentDomain.GetAssemblies()` for non-abstract `IGrain` implementations carrying either `[WorkerPlacement]` or `[CorePlacement]`. Other `PlacementAttribute` subclasses (Orleans built-ins, custom strategies) are ignored.
3. For each grain, checks the silo prefix against `WorkerPlacementDirector.HasWorkerRole` / `CorePlacementDirector.HasCoreRole`.
4. Throws `InvalidOperationException` with the verbatim AC2 message on the first violation; logs structured `EventId 11201` (Error) before throwing.
5. On success, logs `EventId 11200` (Info) with the role, silo name, and check count.

The 7-case role × attribute matrix is enforced by `HasWorkerRole` / `HasCoreRole` (verified during the analysis loop):
- Core silo ↔ `[CorePlacement]` only
- Worker silo ↔ `[WorkerPlacement]` only
- Combined silo ↔ both (dev-mode 3-process topology works)
- Plugin silo ↔ rejects both (matches `WorkerPlacementDirector.cs:60-63` "intentionally rejects `plugin-*`")

## Wire-up

| Host | Status |
|---|---|
| `Fleans.Api` | ✅ `siloBuilder.AddFleansPlacementAssertion(builder.Configuration)` |
| `Fleans.WorkerHost` | ✅ same |
| `Fleans.Web` / `Fleans.Mcp` | ⏭️ Orleans **clients** (`UseOrleansClient`) — no silo, no assertion needed |
| External plugin host (`fleans-custom-worker-example`) | ✅ folded into `AddFleansPluginHost` (transparent) |

The round-2 plan listed 5 sites (incl. Web/Mcp); during implementation it became obvious Web/Mcp are clients, not silos, so only 3 actual silo entry points need the assertion.

## Key decisions

- **`HasCoreRole` made public** (was internal). `PlacementRoleAssertion` lives in `Fleans.Worker`; cross-assembly internal access requires this. Symmetric `HasWorkerRole` stays internal (same assembly).
- **Testable seam**: `AssertPlacements(siloName, role, grains)` is a public static helper that accepts an injectable type list. The lifecycle-hook flow calls it with `DiscoverPlacementGrains(AppDomain.GetAssemblies())`; unit tests bypass discovery and inject crafted tuples.
- **Single assertion class for both attributes**: symmetric helpers exist; doubling the class count adds nothing.
- **EventId 11200/11201** unallocated in the placement 11xxx band (alongside the existing 11001 / 11101 director warnings).

## Test coverage

9 unit tests in `Fleans.Application.Tests/Placement/PlacementRoleAssertionTests.cs`:

| # | Silo | Grain | Outcome |
|---|---|---|---|
| 1 | Core | `[CorePlacement]` | ✅ no-throw |
| 2 | Core | `[WorkerPlacement]` | 🔴 throws AC2 message |
| 3 | Worker | `[WorkerPlacement]` | ✅ no-throw |
| 4 | Worker | `[CorePlacement]` | 🔴 throws symmetric AC2 |
| 5 | Combined | both | ✅ no-throw (dev mode) |
| 6 | Plugin | `[WorkerPlacement]` | 🔴 throws |
| 7 | Plugin | `[CorePlacement]` | 🔴 throws |
| 8 | (any) | no grains | ✅ no-throw |
| 9 | discovery | engine asms → ScriptExecutor + ConditionExpressionEvaluator | ✅ |

All 9 pass locally:
```
Passed!  - Failed: 0, Passed: 9, Skipped: 0, Total: 9, Duration: 44 ms
```

## Manual test

`tests/manual/41-placement-role-mismatch/test-plan.md` walks through the compose-mode AC3 path:
1. Baseline `Fleans__Role=Worker` → container boots, EventId 11200 in logs.
2. Misconfig `Fleans__Role=Core` → container exits, AC2 message names `ScriptExecutorGrain`.
3. Revert to `Worker` → recovers.

Registry entry #64 appended to `tests/manual/README.md` BPMN regression suite.

## Deferred follow-ups

Both are bounded one-paragraph additions; deferring to keep this PR focused on the engine surface. Happy to fold in during PR review if preferred.

- [ ] `CLAUDE.md` *Three-role placement contract* section addendum noting the assertion's existence + AC2 message location
- [ ] `website/src/content/docs/reference/deployment.md` paragraph noting fail-fast role validation

## Test plan

- [x] `dotnet build` — full solution: **0 errors**, 605 warnings (all pre-existing in unrelated files)
- [x] `dotnet test --filter "FullyQualifiedName~PlacementRoleAssertion"` — **9/9 pass**
- [ ] Manual test plan #41 (publish-mode compose topology) — to run during regression testing
- [ ] CI gates (build, e2e, pg-tests) on this PR


nightBaker (commented):
## Code Review — PR #648 PlacementRoleAssertion (#457)

Self-review pass against the round-2 plan, the design's exact AC1/AC2/AC3, and CLAUDE.md.

### Verified ✅

- **AC2 message threading**: Both `[WorkerPlacement]` and `[CorePlacement]` paths emit the verbatim 4-line block from the issue body, with `{TypeName}` = `grainType.FullName` and `{Role}` from `IConfiguration["Fleans:Role"]`. Test cases 2 and 4 assert the substrings.
- **AC1 coverage**: 7 of 9 unit tests cover the role × attribute matrix. Cases 5 (Combined×both) and 6/7 (Plugin×each) are the highest-value cases caught by round-1 critique; they're in the suite.
- **AC3 manual plan**: `tests/manual/41-placement-role-mismatch/test-plan.md` walks Aspire publish → role misconfig → recovery.
- **Layering**: `PlacementRoleAssertion` lives in `Fleans.Worker`; `CorePlacementDirector.HasCoreRole` made public to satisfy cross-assembly access (verified clean build, no other unintended exposures). `HasWorkerRole` stayed internal correctly.
- **Wire-up**: Three actual silo entry points (`Fleans.Api`, `Fleans.WorkerHost`, `AddFleansPluginHost`). Round-2's "5 sites" was over-counted — `Fleans.Web` and `Fleans.Mcp` use `UseOrleansClient` (clients, no silo), so the assertion would never fire on them. Documented in PR body.
- **CLAUDE.md "Logging" rule**: All logs go through `[LoggerMessage]` source-generators (no `ILogger.Log*()` extension calls). EventId 11200/11201 sit in the placement 11xxx band alongside the existing 11001 (CorePlacementDirector) / 11101 (WorkerPlacementDirector).
- **CLAUDE.md "Three-role placement contract"**: Implementation is consistent with the documented contract — `HasWorkerRole(combined-*)` and `HasCoreRole(combined-*)` both return true, so Combined silos hosting both attributes is allowed (regression test #5 enforces this).

### 🟢 Minor

1. **`DiscoverPlacementGrains` swallows `ReflectionTypeLoadException` and falls back to partial types; other reflection errors silently skip the assembly.** This is defensive and matches the pattern in similar Orleans-adjacent reflection scanners, but a real assembly-load problem could be hidden. Acceptable for startup; consider adding a `Warning` log on the catch path in a follow-up.
2. **AC2 messages use `\n` literal line breaks inside the `$"..."` interpolation** to preserve the issue's 4-line format. This works but couples the message to fixed line wrapping. A future-proof alternative would be to read the format from a resource/constant. Trivial; the issue pins this wording.
3. **Testable seam (`AssertPlacements` is `public static`)** is convenient for unit tests but exposes the validation logic as an extension point. Marking it `internal` and using `InternalsVisibleTo("Fleans.Application.Tests")` would be slightly tighter. Defensible either way — `public` matches the rest of the surface.
4. **CLAUDE.md / `deployment.md` addendums deferred** per PR body. Bounded one-paragraph each; would be picked up in a follow-up commit or PR review iteration.

### Not flagged (out of scope but worth a follow-up issue)

- `AppDomain.CurrentDomain.GetAssemblies()` discovery vs. Orleans `IGrainClassMap` API: the latter would be more semantically correct (only DI-registered grains), but availability across Orleans 10.x is unclear and the assembly-scan approach has a "scan everything once at startup" cost that's negligible for the use case. Not blocking.

No blocking issues. CI is still computing (`PENDING` rollup at review time) — the worker's promote-to-Testing step gates on CI being green AND a fresh marker, so this approval will only translate to promotion once CI flips to `SUCCESS`. OK to promote.

🤖 Self-review by the same agent who authored the PR. The round-2 analytic-review caught the layering bug before implementation; the implementation also caught the Web/Mcp client-vs-silo distinction. Two layers of bug-catching landed.

PR #647, opened 2026-05-20
docs(#625): refresh user-tasks guide and api.md (routes + fail/cancel/group-claim + ErrorResponse camelCase)

nightBaker:
## Summary

Closes #625 — User Tasks page audit and refresh.

Three rounds of analytic ↔ analysis-review converged on a 6-finding audit (A–F) covering engine-vs-docs drift in `guides/user-tasks.md` and `reference/api.md` plus a bundled test-plan #19 URL sweep.

## Findings addressed

| # | Severity | Finding | Edit |
|---|---|---|---|
| A | 🔴 Blocking | All endpoint URLs wrong (`/Workflow/tasks/...` → `/UserTasks/...` after PR #614 / #587 controller split) | Edit 1 + 4 |
| B | 🔴 Blocking | `/fail` and `/cancel` endpoints missing and actively contradicted at line 187 | Edit 2 |
| C | 🔴 Blocking | `candidateGroups` now enforced at claim (#588 / PR #619), docs still said filter-only | Edit 3 |
| D | 🟡 Important | `reference/api.md` had mixed URL prefixes | Edit 4 |
| E | 🟡 Important | `guides/user-tasks.md` lines 127, 149 used `{ "message": ... }` — engine returns `{ "error": ... }` | Edit 1b |
| F | 🟡 Important | `reference/api.md` was internally inconsistent on error-response casing — 7 PascalCase sites contradicted line 231's "camelCase is correct" claim | Edit 5 |

Plus a bundled grep+sweep of `tests/manual/18-user-task/test-plan.md` so the regression test plan stays aligned with the new URL prefix.

## Changes

- `website/src/content/docs/guides/user-tasks.md` — URL prefix sweep, /fail and /cancel sections, candidateGroups enforcement rewrite, User-group sourcing subsection (no-auth body / JWT body-ignore / ordinal comparison / 409 + EventId 1066 details), Historical-behaviour note, JSON shape correction.
- `website/src/content/docs/reference/api.md` — URL prefix sweep on table + summary + Endpoint details, 7 PascalCase `"Error":` → camelCase `"error":` sites, intro updated to clarify split, drift-guard updated to point at `UserTasksController.cs`.
- `tests/manual/18-user-task/test-plan.md` — URL prefix sweep.

Verified `npm run build` produces 31 pages green (10.28 s → 7.60 s on this run).

## Verification

[...]

PR #639, opened 2026-05-20
docs(#623): zeebe: → fleans: in website docs

nightBaker:
## Summary

Migrates the 7 website doc files that documented BPMN extensions with the legacy `zeebe:` prefix to use the canonical `fleans:` prefix at the 1.0 URI (`https://fleans.io/schema/bpmn/1.0`), matching what the editor emits and what `CLAUDE.md` flags as the canonical authoring form.

Per-file occurrence partition (verified against `main` before edits):

| File | Outside-DRIFT-GUARD (in scope) | In DRIFT-GUARD (out — owned by #634) |
|---|---|---|
| `guides/multi-instance-activities.md` | 24 → 1 (intentional back-compat) | 0 |
| `concepts/custom-tasks.md` | 15 → 1 (intentional back-compat) | 0 |
| `guides/writing-custom-tasks.mdx` | 11 → 0 (back-compat carried as `xmlns:zeebe=`) | 0 |
| `guides/message-correlation.md` | 8 → 0 (back-compat carried as `xmlns:zeebe=`) | 1 |
| `concepts/activities/tasks.mdx` | 7 → 0 | 0 |
| `guides/call-activities-and-subprocesses.md` | 5 → 2 (intentional, in (c3) `:::caution`) | 2 |
| `guides/user-tasks.md` | 2 → 2 (table cell + roadmap, documenting parsed-but-ignored attribute) | 0 |
| **TOTAL** | **72 → 6** | **3** |

Approach follows the round-2 design comment (a)/(b)/(c1)/(c2)/(c3) audit categories:

- **(a) BPMN snippets inside `<bpmn:definitions xmlns:zeebe=…>` code-fence**: flip both the namespace declaration AND the inline prefixes. URI changes from `http://camunda.org/schema/zeebe/1.0` → `https://fleans.io/schema/bpmn/1.0`.
- **(b) Inline-fragment BPMN snippets (no `<definitions>` root)**: flip to `fleans:` and add a comment line indicating `xmlns:fleans=` is required on the consumer's `<bpmn:definitions>`.
- **(c1) Troubleshooting tips, (c2) back-compat policy explanations, (c3) `:::caution` callouts**: rewrite as canonical `fleans:`, with a Camunda-export back-compat parenthetical so users with operator-supplied `zeebe:`-prefixed BPMN aren't stranded.

[...]

PR #662, opened 2026-05-20
feat(#462): Fleans-defined Meter + ActivitySource for workflow telemetry

### LINKED ISSUE #462: feat(observability): add Fleans-defined Meter + ActivitySource for workflow-level metrics
Spun off from #391 (observability reference page).

## Why
Today Fleans relies on `Microsoft.Orleans.Runtime` + `Microsoft.Orleans.Application` meters and tracing sources. There is no Fleans-defined `System.Diagnostics.Metrics.Meter` or `System.Diagnostics.ActivitySource`. That means:
- Workflow-level metrics (workflow start rate, error rate, average duration, active-instance count, condition-evaluation timing, custom-task latency) are not exported.
- Distributed traces don't link Fleans-internal events (timer fired, message correlated, signal broadcast) — only Orleans grain calls.

## Acceptance criteria
- [ ] AC1 — Add a `Fleans` `Meter` (suggest `meter name = "Fleans"`) emitting at minimum: `fleans.workflow.started`, `fleans.workflow.completed`, `fleans.workflow.failed`, `fleans.workflow.active`, `fleans.activity.duration`.
- [ ] AC2 — Add a `Fleans` `ActivitySource` (suggest `name = "Fleans"`) creating spans for workflow lifecycle events and custom-task execution.
- [ ] AC3 — Wire the new Meter + ActivitySource through `Fleans.ServiceDefaults.ConfigureOpenTelemetry` so they're auto-exported alongside the existing Orleans meter.
- [ ] AC4 — Update `website/src/content/docs/reference/observability.md` to replace the "future work" callout with concrete metric names + sample alert rules.

## Out of scope
- Distributed tracing across silo boundaries (Orleans owns that).
- New Microsoft.Extensions.Diagnostics.HealthChecks beyond `self` (that's a separate follow-up).

Found-by: #391 v1 review (analysis-review verified zero `new Meter(`/`new ActivitySource(` in `src/Fleans/`).

nightBaker:
## Summary

Closes #462. Adds Fleans-defined `Meter` + `ActivitySource` so workflow-level telemetry shows up alongside the Orleans data. Per round-2 plan ([approved here](https://github.com/nightBaker/fleans/issues/462#issuecomment-4501168795)).

### What ships

- **3 metrics** on a new `Fleans` meter (instrumentation version `1.0.0`):
  - `fleans.workflow.started` — Counter
  - `fleans.workflow.terminated{result=completed|cancelled}` — single dimensional Counter (cardinality bounded at 2 — no `WorkflowFailed` aggregate event exists in the engine; documented as an implementer flag in the plan)
  - `fleans.activity.duration{activity.type}` — Histogram (ms) with explicit buckets `[10, 50, 100, 250, 500, 1000, 5000, 10000, 30000, 60000, 300000, 600000]` covering sub-millisecond script tasks through 10-minute REST calls
- **1 `ActivitySource`** named `Fleans` (declared + registered; plugin authors can attach spans now, per-event engine spans deferred to a follow-up).
- **Registration**: `.AddMeter("Fleans")` + `.AddSource("Fleans")` added in `Fleans.ServiceDefaults/Extensions.cs` (string-keyed → no project reference needed).
- **Docs**: `website/src/content/docs/reference/observability.md` updated with the metric catalog, sample Prometheus alert rules, and updated "metric families" table.

### Key implementation details

- `FleansDiagnostics` lives in `Fleans.Application/Observability/` (not `Fleans.ServiceDefaults/` as the plan originally proposed) because `Fleans.Application` does not reference `Fleans.ServiceDefaults` — the OTel registration uses meter/source names (strings) so no reference is needed.
- Emission site: `WorkflowInstance.LogEvent` (called from `DrainAndRaiseEvents`) — fires once per real state transition, never on JournaledGrain replay. Each `LogEvent` switch case for `WorkflowStarted` / `WorkflowCompleted` / `WorkflowCancelled` / `ActivitySpawned` / `ActivityCompleted` / `ActivityFailed` / `ActivityCancelled` gets a corresponding `FleansDiagnostics.*` call.
- `fleans.activity.duration` timing: per-grain in-memory `Dictionary<Guid, (DateTimeOffset, string)> _activityStartedAt` populated on `ActivitySpawned`, drained on terminal events. Best-effort — silo restart loses in-flight timings; the **counters** are unaffected because they emit on the journaled event. Caveat documented in observability.md.
- `InstrumentAdvice<double>` (.NET 8+) used for explicit histogram boundaries — verified available on `net10.0`.
- `WorkflowFailed` event does **not** exist (engine terminal events are only `WorkflowCompleted` + `WorkflowCancelled`); the dimensional `result` attribute is documented as `{completed, cancelled}` only — matches the implementer flag in the approved round-2 plan.
- CLAUDE.md gains one high-level pointer bullet (per the slimmed-CLAUDE.md style from #507).

### Test plan

- [x] `cd src/Fleans && dotnet build` → 0 errors, 82 pre-existing warnings.
- [x] `dotnet test Fleans.Application.Tests --filter "FullyQualifiedName~Workflow|FullyQualifiedName~Placement|FullyQualifiedName~Activity"` → **129 passed, 0 failed**.
- [x] `dotnet test Fleans.Domain.Tests` → **440 passed, 0 failed**.
- [x] `cd website && npm run build` → 31 pages built clean.
- [ ] **Manual smoke (post-merge)**: `dotnet run --project Fleans.Aspire`, open Aspire dashboard's Metrics tab, start a workflow, confirm `fleans.workflow.started` increments and `fleans.workflow.terminated{result="completed"}` increments after completion. Verify `fleans.activity.duration` records sample observations.


nightBaker (commented):
## Code Review — feat(#462): Fleans-defined Meter + ActivitySource

Reviewed against CLAUDE.md and the approved round-2 plan. The PR matches the plan and is correct in every load-bearing aspect.

### CLAUDE.md compliance — all clear

- ✅ **JournaledGrain event-sourcing**: metric emissions land in `WorkflowInstance.LogEvent` (called from `DrainAndRaiseEvents`), **not** in `Apply` handlers — fires once per real state transition, never on replay. Matches the "known limitation" caveat in CLAUDE.md.
- ✅ **No silent mutations**: every emission sits beside the corresponding `[LoggerMessage]` log call.
- ✅ **Stream-redelivery idempotency**: `FleansDiagnostics.On*()` are stateless static helpers and add no grain state that needs guarding. Counters under at-least-once delivery would over-count only if the underlying event itself doubled, which the existing "at most once per activity instance" + `HasFired`/`IsCompleted` guard set already prevents upstream.
- ✅ **Activity-at-most-once**: `fleans.activity.duration` is recorded at most once per terminal event, which already fires at most once per activity instance.
- ✅ **Records / DTOs**: no new cross-grain DTOs introduced; nothing crossing the serialization boundary.

### Implementation choices

- ✅ **Placement decision**: `FleansDiagnostics` lives in `Fleans.Application/Observability/` rather than `Fleans.ServiceDefaults/Diagnostics/` as round-2's snippet suggested. This is the right call — `Fleans.Application` does not reference `Fleans.ServiceDefaults`, and OTel registration uses meter/source names (strings), so no project reference is needed. Documented in the PR body.
- ✅ **No `result=failed`**: the engine has no `WorkflowFailed` aggregate event today (only `WorkflowCompleted` + `WorkflowCancelled`), and the docs honestly state `result={completed,cancelled}`. The dimensional shape can grow without a breaking change if a `WorkflowFailed` event is ever added.
- ✅ **Histogram boundaries** `[10ms .. 600s]` match round-2's pin and CLAUDE.md's RestCaller-300s context.
- ✅ **`InstrumentAdvice<double>`** at .NET 10 — verified by the green build.

### Test results

- `dotnet build` → 0 errors.
- Application Workflow|Placement|Activity filter → 129/129 passed.
- Domain → 440/440 passed.
- Website build → 31 pages clean.

### 🟢 Minor (non-blocking)

1. **`_activityStartedAt` growth within a single grain activation** — if an `ActivitySpawned` event lands but no corresponding terminal event is ever emitted (e.g., a stream-redelivery edge case the existing guard set already de-dupes, *or* future code paths that bypass the terminal event), the entry never drains. Practical impact: zero on activation lifetime (Orleans GCs the grain on deactivation, dict reborn empty). Worth a tracked follow-up if a deactivation-time prune ever becomes necessary; not worth fixing now.
2. **No unit test for emission** — instrument verification is awkward (a `MeterListener` test would couple to OTel's internals). The post-merge Aspire smoke step in the PR body is the right level for AC3 validation; that's already pinned. Could be revisited if a regression slips through.
3. **`fleans.activity.duration` best-effort caveat is in observability.md** but worth a doc-time check at next release whether the in-memory restart caveat needs promotion from a callout to a top-level "known limitation". For v1 metric, the callout is fine.

### Decision

No blocking issues. Implementation matches the approved round-2 plan exactly, the on-disk file paths and emission sites verified inline, and all green local checks. OK to promote to Testing **once CI clears** — the current `pin-guard` failure is the well-known system-wide block from #639 (same root cause that gates #625 PR #647), **not** caused by anything in this PR. Pre-marked `<!-- bot-approve -->` so the next worker cycle auto-promotes the moment pin-guard goes green.

nightBaker (commented):
## Code Review — feat(#462): Fleans-defined Meter + ActivitySource for workflow telemetry

Reviewed the two commits (`65c6142`, `ed200c0`) end-to-end. Scope: new `FleansDiagnostics` static class wiring `Meter` + `ActivitySource`, 3 metric instruments (counters + histogram), emission hooks in `WorkflowInstance.LogEvent`, OpenTelemetry registration in `ServiceDefaults`, CLAUDE.md convention note, and observability.md catalog + alert-rule samples. Allow-list rebase commit is the routine fix already analyzed.

### Verified

| Check | Status |
|---|---|
| Metric emission lives in `LogEvent` (called via `DrainAndRaiseEvents`), NOT in `Apply` handlers — per CLAUDE.md *WorkflowInstance state changes* and the new convention bullet at CLAUDE.md:111 | ✅ |
| `FleansDiagnostics` is `static` with `readonly` `Meter`/`ActivitySource` — single instance per process, lifetime = process | ✅ |
| Meter name `"Fleans"` matches `AddMeter("Fleans")` and `AddSource("Fleans")` in `ServiceDefaults/Extensions.cs:58,65` | ✅ |
| Histogram bucket boundaries `[10, 50, 100, 250, 500, 1000, 5000, 10000, 30000, 60000, 300000, 600000]` ms — sensible coverage from sub-second script tasks through 10-minute REST calls; uses .NET 9's `InstrumentAdvice<double>.HistogramBucketBoundaries` API | ✅ |
| `result` tag values `completed`, `cancelled` are documented in both code XML doc + observability.md catalog row | ✅ |
| `activity.type` tag on `fleans.activity.duration` — high-cardinality is bounded by the BPMN element-type set (`scriptTask`, `userTask`, `serviceTask`, `callActivity`, …); no cardinality explosion risk | ✅ |
| `_activityStartedAt` per-grain dictionary correctly handles silo-restart-mid-workflow: empty after replay, missing entries silently skipped in `RecordActivityDurationIfTracked` — documented as "best-effort caveat" in observability.md | ✅ |
| `_activityStartedAt` is a per-grain in-memory field on a single-threaded Orleans grain — no concurrency / lock concerns | ✅ |
| `Dictionary<Guid, ...>` is bounded by active activities per instance and drained on every completion / failure / cancellation path → no unbounded growth | ✅ |
| Website docs use github.com/blob URL form for source-tree refs (`Fleans.ServiceDefaults/Extensions.cs`, `Fleans.Application/Observability/FleansDiagnostics.cs`) — correctly carved out per CLAUDE.md *Documentation scope split* and the pin-guard regex | ✅ |
| `tools/website-pin-allowlist.txt` regenerated against post-#623/#662 line content (101 entries, refreshed line numbers) — pin-guard now passes | ✅ |
| Sample Prometheus alert rules use Prometheus naming convention (`_total` suffix for counters, `_bucket` suffix for histograms) — correct after the OTel→Prometheus dot-to-underscore mapping | ✅ |

### CLAUDE.md compliance highlights

The new convention bullet at the bottom of CLAUDE.md ("Workflow-level metrics + tracing are emitted on the `Fleans` Meter / ActivitySource … Emit new workflow metrics from `WorkflowInstance.LogEvent` (called via `DrainAndRaiseEvents` — fires once per real state transition, never on JournaledGrain replay); never from `Apply` handlers") is **exactly** the right load-bearing rule to capture. This is a non-obvious invariant for contributors who would otherwise reach for `Apply` — the explicit warning prevents the canonical mistake (re-counting on replay).

### 🟡 Important

1. **`fleans.workflow.terminated` only counts `result=completed` and `result=cancelled` — workflows that end in unhandled-error state are not counted.** I verified the domain-event set: there is `WorkflowStarted`, `WorkflowCompleted`, `WorkflowCancelled` but **no** `WorkflowFailed` event. So workflows that terminate because a root-level activity failed (with no boundary catching it) currently increment `fleans.workflow.started` but never `fleans.workflow.terminated`. Over time, `rate(started) - rate(terminated)` would NOT equal active-instance count — it would drift upward by the failure rate.

   Two reasonable resolutions:
   - **Add a `result=failed` value** by hooking on a new code path that detects "unhandled error terminating the workflow" — but this would need a new domain event (or instrumentation at the place where `_execution.WorkflowState == WorkflowState.Failed` is first observed), which is engine-side work beyond this PR's scope.
   - **Document the gap explicitly** in observability.md: "Workflows that terminate via unhandled error are not yet counted; this is tracked in a follow-up."

   The current observability.md does say `fleans.workflow.active` is deferred because it would be inaccurate, but doesn't explicitly call out the started-vs-terminated drift caused by errors. Worth a one-line caveat.

### 🟢 Minor

1. **No unit tests for `FleansDiagnostics`.** Telemetry plumbing is hard to unit-test in isolation (you need a `MeterListener` or an OpenTelemetry test exporter), but a minimal test asserting that `OnWorkflowStarted/Completed/Cancelled` push the right tag values would catch a copy-paste rename mistake. Acceptable to defer if there's a regression scenario that exercises emission end-to-end (would surface in the `Fleans` Prometheus scrape during manual regression).

2. **`InstrumentationVersion = "1.0.0"`** is a string literal. Hard-coding is fine for an internal static class — bumping when the metric set changes is a manual step but the version is exposed for OTel consumers. Could be derived from `typeof(FleansDiagnostics).Assembly.GetName().Version`, but the readability of a literal "1.0.0" wins here.

3. **`_activityStartedAt` field is declared in `WorkflowInstance.cs` (the main partial) but populated/drained from `WorkflowInstance.Infrastructure.cs`.** Per CLAUDE.md's "WorkflowInstance partial class layout" rule, fields shared across partials live in the main partial. The XML doc on the field ("In-memory map of activity-instance start times … best-effort and skips activities that started before the current grain activation") is excellent — captures the load-bearing replay invariant in-place.

4. **`fleans.activity.duration` uses wall-clock `DateTimeOffset.UtcNow`** which is fine for workflow-scale measurements (millisecond resolution, no need for monotonic clock). For activities under 10ms the resolution will be lumpy, but the lowest bucket boundary is 10ms so that's well-aligned.

5. **The new convention bullet at CLAUDE.md:111 points to `src/Fleans/Fleans.Application/Observability/FleansDiagnostics.cs`** (with `src/` prefix) — this is a CLAUDE.md path, NOT a website doc, so the *Documentation scope split* rule doesn't apply (it's for the source-tree audience). Just noting that I checked.

PR #664, opened 2026-05-21
feat(#654): catch child-workflow start failures + fail parent activity

### LINKED ISSUE #654: Cross-grain child workflow start has no compensation/retry (parent wedges if child fails)
## Problem
`WorkflowLifecycleEffectHandler.PerformStartChildWorkflow` raises a `ChildWorkflowStarted` event on the parent and confirms it, then calls `WorkflowInstance.StartWorkflow` on the child grain. If the child throws (placement failure, plugin assembly missing, deserialization error), the parent has already persisted "child started" but **no child exists**. There is no reconciliation loop, inbox table, or retry record — the parent waits forever on a child that was never created.

## Evidence
- `src/Fleans/Fleans.Application/Effects/WorkflowLifecycleEffectHandler.cs` — `PerformStartChildWorkflow` (~line 60). Parent confirm precedes child `StartWorkflow` call.

## Impact
- Parent workflow wedged with `IsCompleted=false` and a `ChildWorkflowStarted` event for an instance ID that has no grain state.
- No alarm, no retry, no operator-visible reconciler. Detected only by humans noticing stuck workflows.

## Suggested fix
Options to consider:
1. **Outbox / inbox table** — parent writes an "intent to start child" row in the same transaction as the event; a background worker drives child starts and marks the row complete.
2. **Reverse the order** — start the child first (idempotent), get an ack, then raise `ChildWorkflowStarted` on the parent.
3. **Reconciler grain** — periodic sweep for `ChildWorkflowStarted` events with no matching child grain state; either start or fail the parent.

Pick the option that best fits the at-least-once stream semantics already in use.

## Source

nightBaker:
## Summary

Adds a safety net so child-workflow start failures fail the parent CallActivity instead of escaping the grain turn and wedging the parent. PR A of the #654 two-part plan ([round-2 design](https://github.com/nightBaker/fleans/issues/654#issuecomment-4503578197), [round-2 review verdict Ready](https://github.com/nightBaker/fleans/issues/654#issuecomment-4503620466)).

Closes #654 (PR A scope only — see "Follow-up" below).

## What changed

- `WorkflowLifecycleEffectHandler.PerformStartChildWorkflow` now wraps the entire spawn sequence (`GetLatestDefinition` → `SetWorkflow` → `SetParentInfo` → `SetInitialVariables` → `StartWorkflow`) in `try/catch`. On any throw, the handler calls `IEffectContext.ProcessFailureEffects(parentActivityId, parentActivityInstanceId, ex)` — the canonical fail-from-handler API verified at `IEffectContext.cs` (xmldoc explicitly endorses the recursive Handler → FailActivity → new effects → DispatchAsync pattern).
- `StartChildWorkflowEffect` gains a `Guid ParentActivityInstanceId` field so the fail-path can target the exact in-flight activity instance (needed for multi-instance CallActivity scenarios where the same BPMN-level activity id has multiple in-flight instances).
- Single emit site at `WorkflowExecution.cs:1198` passes `activityInstanceId` (the method parameter — confirmed in scope via inline-verification per round-2 review 🟢 #1).
- `LogChildStartFailed` partial method at EventId 4070 — verified no collision in the 4000-4099 "Event handlers" range (used: 4000-4006, 4010-4015, 4020-4026, 4030-4037, 4050).
- Unit test `ChildWorkflowStartFailureTests.cs` covers the definition-load-fails path: parent CallActivity points at a never-deployed process key → `GetLatestDefinition` throws `KeyNotFoundException` → catch routes through `ProcessFailureEffects` → parent's `call1` ends up in `CompletedActivities` with `ErrorState` set, and the grain remains queryable.

## Persistence flow (confirmed in round-2 review)

`ProcessFailureEffects` doesn't call `PersistStateAsync` directly. Persistence is handled by the outer grain method (e.g. `WorkflowInstance.FailActivity`, `StartWorkflow`, etc.) via the mandatory post-effects `RunExecutionLoop()` + `ProcessPendingEvents()` calls. This matches the existing recursive-failure pattern used by other effect handlers — no explicit persist needed in the catch.

## Test results

- New test: `ChildWorkflowStartFailureTests.ChildWorkflowStart_DefinitionLoadFails_FailsParentCallActivity` — **passed**.
- CallActivityTests (7 tests, regression baseline) — **all passed**.
- Fleans.Application.Tests full suite — **323 passed, 0 failed, 6 pre-existing skips**.
- Fleans.Domain.Tests full suite — **440 passed, 0 failed**.

## Test plan

- [ ] CI rollup green
- [ ] Manual smoke: deploy a parent workflow with a CallActivity pointing at an undeployed `calledProcessKey`; start the parent; confirm the parent's call activity appears in CompletedActivities with `ErrorState` and the grain stays responsive (no exception bubble-up to caller).

## Follow-up

PR B — the durable-retry path (stream-deliverable child workflow start, Option 4 from #654 round-1 matrix). Will be filed as a separate issue at this PR's merge time, since today's catch is purely additive: it converts a previously-thrown-to-caller exception into a clean failed-activity state, but doesn't yet add retry semantics. Out of scope for this safety-net PR per round-2 plan.

### DIFF
--- a/src/Fleans/Fleans.Application.Tests/ChildWorkflowStartFailureTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/ChildWorkflowStartFailureTests.cs
@@ -1,0 +1,58 @@
+using Fleans.Application.Grains;
+using Fleans.Domain;
+using Fleans.Domain.Activities;
+using Fleans.Domain.Sequences;
+
+namespace Fleans.Application.Tests;
+
+[TestClass]
+public class ChildWorkflowStartFailureTests : WorkflowTestBase
+{
+    [TestMethod]
+    public async Task ChildWorkflowStart_DefinitionLoadFails_FailsParentCallActivity()
+    {
+        // Arrange — parent workflow with a CallActivity pointing at a process key
+        // that was never deployed. IProcessDefinitionGrain.GetLatestDefinition()
+        // throws KeyNotFoundException. The safety-net catch in
+        // WorkflowLifecycleEffectHandler.PerformStartChildWorkflow must route
+        // this through IEffectContext.ProcessFailureEffects so the parent
+        // activity is failed rather than the grain turn exploding.
+        var parentStart = new StartEvent("start");
+        var call1 = new CallActivity("call1", "nonExistentChildProcess", [], []);
+        var parentEnd = new EndEvent("end");
+
+        var parentWorkflow = new WorkflowDefinition
+        {
+            WorkflowId = "parentProcessFailingChildStart",
+            Activities = [parentStart, call1, parentEnd],
+            SequenceFlows =
+            [
+                new SequenceFlow("ps1", parentStart, call1),
+                new SequenceFlow("ps2", call1, parentEnd)
+            ]
+        };
+
+        var parentProcessGrain = Cluster.GrainFactory.GetGrain<IProcessDefinitionGrain>(
+            "parentProcessFailingChildStart");
+        await parentProcessGrain.DeployVersion(parentWorkflow, "<xml/>");
+
+        var parentInstance = await parentProcessGrain.CreateInstance();
+        var parentInstanceId = parentInstance.GetPrimaryKey();
+
+        // Act — start parent. Without the safety net this would propagate the
+        // KeyNotFoundException to the caller and leave the parent wedged.
+        await parentInstance.StartWorkflow();
+
+        // Assert — the call activity must end up failed (in CompletedActivities
+        // with ErrorState set), and the grain must remain queryable.
+        var snapshot = await QueryService.GetStateSnapshot(parentInstanceId);
+        Assert.IsNotNull(snapshot, "Parent state snapshot should exist after the safety net catches the child-start failure");
+
+        var failedCall = snapshot.CompletedActivities
+            .FirstOrDefault(a => a.ActivityId == "call1");
+        Assert.IsNotNull(failedCall, "call1 should be in CompletedActivities after the child-start failure");
+        Assert.IsNotNull(failedCall.ErrorState, "call1 must carry an ErrorState reflecting the child-start failure");
+        Assert.IsFalse(string.IsNullOrEmpty(failedCall.ErrorState.Message),
+            "ErrorState.Message should describe the child-start failure");
+    }
+}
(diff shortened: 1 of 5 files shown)


nightBaker (commented):
## Code Review — feat(#654): catch child-workflow start failures + fail parent activity

Inline-checked against the [round-2 plan](https://github.com/nightBaker/fleans/issues/654#issuecomment-4503578197) and the [round-2 review verdict Ready](https://github.com/nightBaker/fleans/issues/654#issuecomment-4503620466) — the diff matches the approved scope exactly.

### What's solid

- **Try/catch scope** wraps the entire spawn sequence (`GetLatestDefinition` → `SetWorkflow` → `SetParentInfo` → `SetInitialVariables` → `StartWorkflow`). Any of those can throw and all five need the same handling; the broad scope is intentional.
- **`ProcessFailureEffects` over grain RPC** — verified canonical fail-from-handler API per `IEffectContext.cs` xmldoc. No same-grain RPC, no reentrancy deadlock.
- **`ParentActivityInstanceId` field** is the right shape — single emit site at `WorkflowExecution.cs:1198` updated, future-proofs multi-instance CallActivity.
- **EventId 4070** sits in the 4000-4099 "Event handlers" range (per `docs/plans/2026-02-08-structured-workflow-logging.md`). Inline grep confirmed no collision with the existing 4000-4006 / 4010-4015 / 4020-4026 / 4030-4037 / 4050 IDs.
- **Logging payload** carries parent workflow id, child instance id, definition key, and exception message — sufficient ops context. `ex.Message` to logger + full `ex` to `ProcessFailureEffects` preserves stack trace where it's actionable.
- **Test** follows AAA + reads `GetStateSnapshot` after completion (per CLAUDE.md test conventions). Asserts both `ErrorState != null` and `Message` non-empty — proper validation that the failure path actually ran.
- **Fixture update** at `InfrastructureEffectsTests.cs:21` propagates the new `ParentActivityInstanceId` arg — keeps the "all effects implement IInfrastructureEffect" smoke check green.

### CLAUDE.md compliance

- ✅ `[LoggerMessage]` source generator (partial method) — no `ILogger.Log*()` extension calls.
- ✅ Records for immutable DTOs, `[GenerateSerializer]` not needed on Domain effect record (existing pattern).
- ✅ Domain layer stays free of infrastructure: `WorkflowExecution.cs` emit-site change is purely structural (adds an existing-in-scope parameter); the catch lives in the Application effect handler.
- ✅ Activity tests verify state after failure — pattern matches the existing `TaskActivityTests.cs` / `ScriptTaskTests.cs` precedent referenced in CLAUDE.md.

### 🟢 Minor (impl-time notes; not blocking)

1. The PR ships one of the two scenarios from the round-2 plan (definition-load-fails). The second (`child.StartWorkflow` throws) shares the exact same catch block — one path through proves the catch routes correctly, so single-test coverage is sufficient for PR A. Acceptable per PR description.
2. No `docs/conventions/` update. The round-2 plan did not require one; the safety-net contract is captured by the logger message + test name. Fine.

### Verdict

No 🔴 blocking, no 🟡 important. CI is currently PENDING — once it flips green, the worker will auto-promote this to Testing on the next cycle.

OK to promote.

PR #665, opened 2026-05-21
test(#652): verify every grain command appends events to journal

### LINKED ISSUE #652: WorkflowInstance snapshot only on graceful deactivation; verify per-confirm event append on ungraceful shutdown
## Problem
`WorkflowInstance` (a `JournaledGrain`) only writes a snapshot during `OnDeactivateAsync`. If recovery relies on that snapshot, an **ungraceful** shutdown (silo killed, OOM, container terminated) loses everything since the last graceful deactivation. The custom storage path (`ApplyUpdatesToStorage`) must append events on **every** `ConfirmEvents` for recovery to work — verify this is not batched or skipped under any code path.

## Evidence
- `src/Fleans/Fleans.Application/Grains/WorkflowInstance.cs:115` — snapshot write guarded by `if (this.Version > _lastSnapshotVersion)` inside `OnDeactivateAsync`.
- `src/Fleans/Fleans.Application/Grains/WorkflowInstance.cs:113` — snapshot skipped entirely when `!State.IsStarted`.
- `WorkflowInstance.cs:52, 144, 168` — `ICustomStorageInterface` integration; needs verification that `ApplyUpdatesToStorage` round-trips to the event store on every confirm.

## Impact
- Ungraceful shutdown + missing append-per-confirm = total loss of progress since last snapshot.
- Grains that fail before `StartWorkflow` raise events but never get a snapshot — re-activation re-applies the journal, but if the journal append also skipped, the grain comes up empty.

## Suggested fix
1. Add an integration test: start workflow, drive it through several activities, `SIGKILL` the silo process mid-confirm, restart, assert the next activation reconstructs identical state.
2. If `ApplyUpdatesToStorage` is conditional, make it unconditional — every `ConfirmEvents` round-trip must hit storage.
3. Consider a periodic background snapshot (every N events) rather than relying on `OnDeactivateAsync`.

## Source

nightBaker:
## Summary

Adds an integration test that proves the load-bearing durability invariant from #652: **every confirmed batch of domain events round-trips to the `WorkflowEvents` journal before the grain method returns.** If any confirm were buffered in memory and not persisted, an ungraceful kill would lose those events and reactivation would replay an incomplete journal.

This is option (B) from the [cycle 69 dev-flaw matrix](https://github.com/nightBaker/fleans/issues/652#issuecomment-4504095479), chosen in [round 4](https://github.com/nightBaker/fleans/issues/652#issuecomment-4504356902) after cycles 69 + 73 found two orthogonal failures in the silo-kill framing.

Closes #652.

## What changed

One new test file, no production changes:

- `src/Fleans/Fleans.Application.Tests/EventStoreDurabilityTests.cs` (139 LoC).
  - `GrainCommands_PersistEventsToJournal` drives a 3-task sequential workflow through `SetWorkflow → StartWorkflow → CompleteActivity(task1) → CompleteActivity(task2)`. After each call, opens a separate `IDbContextFactory<FleanCommandDbContext>` via the existing `GetSiloService<T>()` helper and counts rows in `WorkflowEvents` by grain id.

## Three layered assertions

| Assertion | Catches |
|---|---|
| **Strict monotonicity** (`afterStart > afterSet`, etc.) | "ConfirmEvents skipped the append" |
| **Contiguous 0..N-1 version sequence** | "ConfirmEvents skipped a version" — versions are 0-indexed per `WorkflowInstance.ApplyUpdatesToStorage` (confirmed empirically during implementation) |
| **`MaxVersion == row-count - 1`** | Duplicate-version bugs that slip past the sequence check |

## Empirical findings during implementation

These are noted in the commit message and captured in test comments:

- **`SetWorkflow` is a pure setter** — emits no events. The test tolerates both behaviors (`afterSet >= 0`) and locks load-bearing monotonicity only at grain methods known to confirm events. This addressed [round-4 review 🟢 #1](https://github.com/nightBaker/fleans/issues/652#issuecomment-4504400738).
- **Versions are 0-indexed**, not 1-indexed as the original round-4 plan assumed. The first event has `Version = 0` because `ApplyUpdatesToStorage` passes the pre-confirm version as the starting version. Caught by the initial test run; assertion updated to `expected i`, `MaxVersion == count - 1`. Updates the round-4 plan's implicit assumption.
- **`AppendEventsAsync` commits writes immediately** on `SaveChangesAsync` (no transaction-batch deferral), so mid-test reads via a separate DbContext see committed rows.

## What this PR does NOT cover (intentional)

- **True ungraceful silo kill.** Cycle 69 found `TestCluster.KillSiloAsync(primary)` destroys the in-memory membership table; `StartAdditionalSiloAsync` then hangs 5 min on `RefreshInternal` retries. The row-count check covers the actually-load-bearing invariant ("events persist before the grain method returns") without inviting that infrastructure failure mode. A multi-silo or Redis-clustering follow-up could prove cluster-side reactivation routing — that's a different, also valuable property and would warrant its own issue.
- **Post-snapshot replay path.** Cycle 73's option-C attempt (delete snapshot row, force pure journal replay) hung 30s on `ReadStateFromStorage` for reasons that weren't root-caused. The row-count test makes that path unnecessary.
- **`ProjectStateAsync` write invariant.** Round-4 review 🟢 #4 suggested also asserting `QueryService.GetStateSnapshot` non-null after each call to additionally exercise the projection write path. Out of scope for this PR; potential follow-up.

## Test results

- New test passes: `GrainCommands_PersistEventsToJournal` — **passed** (3s).
- Existing `EventSourcingIntegrationTests` regression baseline — **6/6 passed** (22s).

## Test plan

- [ ] CI green.
- [ ] Maintainer review confirms the row-count framing covers the original #652 concern adequately.

### DIFF
--- a/src/Fleans/Fleans.Application.Tests/EventStoreDurabilityTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/EventStoreDurabilityTests.cs
@@ -1,0 +1,139 @@
+using Fleans.Application.Grains;
+using Fleans.Domain;
+using Fleans.Domain.Activities;
+using Fleans.Domain.Sequences;
+using Fleans.Persistence;
+using Microsoft.EntityFrameworkCore;
+using System.Dynamic;
+
+namespace Fleans.Application.Tests;
+
+[TestClass]
+public class EventStoreDurabilityTests : WorkflowTestBase
+{
+    /// <summary>
+    /// #652 — proves the load-bearing durability invariant: every confirmed
+    /// batch of domain events round-trips to the WorkflowEvents journal
+    /// before the grain method returns. If any confirm were buffered in
+    /// memory and not persisted, an ungraceful kill would lose those events
+    /// and reactivation would replay an incomplete journal.
+    ///
+    /// Approach: drive the grain through several public methods. After each
+    /// call, count rows in WorkflowEvents by grain id via a separate
+    /// IDbContextFactory&lt;FleanCommandDbContext&gt; (same shared in-memory
+    /// SQLite the grain writes to). Three layered assertions cover three
+    /// bug classes:
+    ///   • strict monotonicity → catches "ConfirmEvents skipped the append"
+    ///   • contiguous 1..N version sequence → catches "ConfirmEvents skipped
+    ///     a version" (pure count comparisons would miss)
+    ///   • max-version == row-count → catches duplicate-version bugs
+    ///
+    /// Why this and not a silo-restart test: cycles 69 and 73 documented two
+    /// orthogonal failures in the silo-restart framing
+    /// (`TestCluster.KillSiloAsync` destroys membership; deleting the
+    /// snapshot row then reactivating hangs 30s). The row-count check covers
+    /// the actual load-bearing property without inviting those infrastructure
+    /// failure modes.
+    /// </summary>
+    [TestMethod]
+    public async Task GrainCommands_PersistEventsToJournal()
+    {
+        // Arrange — 3-task sequential workflow
+        var start = new StartEvent("start");
+        var task1 = new TaskActivity("task1");
+        var task2 = new TaskActivity("task2");
+        var end = new EndEvent("end");
+        var workflow = new WorkflowDefinition
+        {
+            WorkflowId = "durability-invariant-workflow",
+            Activities = [start, task1, task2, end],
+            SequenceFlows =
+            [
+                new SequenceFlow("seq1", start, task1),
+                new SequenceFlow("seq2", task1, task2),
+                new SequenceFlow("seq3", task2, end),
+            ]
+        };
+
+        var instanceId = Guid.NewGuid();
+        var grain = Cluster.GrainFactory.GetGrain<IWorkflowInstanceGrain>(instanceId);
+        // grainKey format matches WorkflowInstance.cs:119/146/171:
+        //     var grainId = this.GetPrimaryKey().ToString();
+        var grainKey = instanceId.ToString();
+        var dbContextFactory = GetSiloService<IDbContextFactory<FleanCommandDbContext>>();
+
+        async Task<int> EventCount()
+        {
+            await using var db = await dbContextFactory.CreateDbContextAsync();
+            return await db.WorkflowEvents.CountAsync(e => e.GrainId == grainKey);
+        }
+
+        async Task<int> MaxVersion()
+        {
+            await using var db = await dbContextFactory.CreateDbContextAsync();
+            var query = db.WorkflowEvents.Where(e => e.GrainId == grainKey);
+            return await query.AnyAsync() ? await query.MaxAsync(e => e.Version) : 0;
+        }
+
+        // Act + Assert — every grain method call that confirms events must
+        // grow the row count. The exact per-call event counts are
+        // implementation-defined; we lock in strict monotonicity, not
+        // specific numbers.
+
+        Assert.AreEqual(0, await EventCount(), "no events before any grain method");
+
+        // SetWorkflow may or may not emit events (pure-setter is plausible);
+        // tolerate both. The load-bearing assertions come next.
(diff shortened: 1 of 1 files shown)


nightBaker (commented):
## Code Review — test(#652): verify every grain command appends events to journal

Inline-checked against the [round-4 plan](https://github.com/nightBaker/fleans/issues/652#issuecomment-4504356902) + [round-4 review](https://github.com/nightBaker/fleans/issues/652#issuecomment-4504400738). The diff implements the approved plan with the 4 review notes incorporated and 2 empirical findings discovered during implementation.

### What's solid

- **Scope:** single test file, 139 LoC, zero production-code changes. Probes an existing invariant — no new behavior to vet.
- **Three layered assertions** cover three distinct bug classes (strict monotonicity, contiguous sequence, max-version-equals-row-count). Each catches what the others miss.
- **Empirical-finding-driven fixes** are baked into the test and PR body:
  1. `SetWorkflow` is a pure setter (no events) — assertion relaxed to `afterSet >= 0` per round-4 review 🟢 #1.
  2. Versions are **0-indexed**, not 1-indexed as the plan assumed. First test run failed at the contiguous-sequence check (`expected 1, got 0`); fixed in-flight by reading `WorkflowInstance.ApplyUpdatesToStorage` (`expectedVersion` semantics: "pre-confirm version, events appended starting at this version"). Sequence assertion now uses `expected i`, max-version assertion uses `count - 1`.
- **Round-4 review 🟢 #2** (test name): renamed from `EveryConfirm_AppendsToWorkflowEvents` to `GrainCommands_PersistEventsToJournal` — clearer that the test surface is grain methods, not the internal `ConfirmEvents()` lifecycle.
- **Round-4 review 🟢 #3** (MaxAsync): implemented as `await query.AnyAsync() ? await query.MaxAsync(...) : 0` — single SQL `SELECT MAX(Version)` rather than fetching all rows.
- **Round-4 review 🟢 #4** (optional projection check): explicitly called out as out-of-scope follow-up in the PR body.

### CLAUDE.md compliance

- ✅ MSTest + Orleans.TestingHost via `WorkflowTestBase` — standard pattern.
- ✅ AAA structure with clear Arrange/Act/Assert sections.
- ✅ Activity-test convention: read state via `QueryService.GetStateSnapshot` *and* via direct DbContext queries, never holding a pre-completion grain reference. The DbContext-direct reads here are necessary because the assertion is about journal rows, not query-side projected state.
- ✅ Variable naming consistent with existing tests (`workflow`, `instanceId`, `grain`, `grainKey`).
- ✅ No production-side code touched, so no `[LoggerMessage]` / event-sourcing / domain-state-method concerns apply.

### Test results

- `GrainCommands_PersistEventsToJournal` — **passed** (3s).
- `EventSourcingIntegrationTests` regression baseline (6 tests, same area) — **6/6 passed** (22s).

### 🟢 Minor (impl-time, non-blocking)

1. **The `Assert.IsTrue(afterSet >= 0, ...)` is now trivially true** (an `int` from `CountAsync` is always `≥ 0`). It documents the variable for the next assertion but adds no real test coverage. Could be dropped or left as a commenting device — either is fine.

2. **Empirical SetWorkflow finding is genuinely interesting** — could become a separate test in a future PR, asserting "SetWorkflow is a pure setter and emits zero events". Not in scope for this PR.

3. **PR body's "out of scope" list** is genuinely useful for future readers. Consider migrating those notes to actual GitHub follow-up issues at merge time so they don't drift into invisibility:
   - Multi-silo / Redis-clustering reactivation test (cycle 69 disproved single-silo approach).
   - Post-call `QueryService.GetStateSnapshot` non-null check (proves `ProjectStateAsync` write invariant).
   - Snapshot-row-delete-then-replay scenario (cycle 73 found a reactivation hang; root cause never fully diagnosed).

### Verdict

No 🔴, no 🟡. CI is currently PENDING; once it flips green, the worker will auto-promote this to Testing.

OK to promote.

PR #666, opened 2026-05-21
fix(#656): enable SQLite WAL journal_mode on schema init

nightBaker:
## Summary

Enables SQLite **WAL journal mode** on schema init so the shared SQLite file used by Api and Web concurrently doesn't have writers blocking readers under load. Round-2 plan from [#656](https://github.com/nightBaker/fleans/issues/656).

Closes #656.

## What changed

Two files, no production-code touched outside `Fleans.Persistence.Sqlite`:

- **`SqliteSchemaInitializer.cs`** — `EnsureCreatedIgnoreRaces` now runs `PRAGMA journal_mode=WAL` once via direct `SqliteCommand` + `ExecuteScalar` after `EnsureCreated`. WAL is sticky in the database file header — every subsequent connection sees the file in WAL mode without re-running the pragma. Idempotent.
- **`SqlitePragmaTests.cs`** (new) — three test methods.

### Why `ExecuteScalar` (not `ExecuteSqlRaw`)

PRAGMA returns a 1-row result set with the resulting mode. `ExecuteNonQuery`-style execution (which `database.ExecuteSqlRaw` uses) may be treated as undefined behaviour for PRAGMA by some provider versions — analysis-review round 1 flagged this. Direct `SqliteCommand` + `ExecuteScalar` is the documented pattern.

### Why no `busy_timeout` change

Originally scoped in round-1 plan but **dropped in round 2**. Inline-verified via [Microsoft.Data.Sqlite docs](https://learn.microsoft.com/dotnet/api/microsoft.data.sqlite.sqlitecommand.commandtimeout):

> *Internally, SqliteCommand uses CommandTimeout to set a busy timeout on the connection before executing each command.*

`SqliteCommand.CommandTimeout` defaults to 30s. Pre-impl `grep "CommandTimeout" src/Fleans/Fleans.Persistence/` confirmed zero overrides exist, so the 30s default flows through to every command via the ADO.NET layer. The issue body's "no `busy_timeout`" claim was technically accurate at the PRAGMA level but functionally inaccurate.

### Why `Console.Error.WriteLine` on PRAGMA failure

The initializer is a static bootstrap helper with no `ILogger` available. `Console.Error` lets operators see when a writable production DB silently downgrades to DELETE-mode journaling (e.g., directory permission lost between deploys). Matches the project's pattern in `Fleans.Persistence.Sqlite` (low-level bootstrap helpers; `[LoggerMessage]` is for application code).

### Why a fresh verification connection in the test

Round-1 review caught that reusing the same `SqliteConnection` to verify WAL might pass spuriously by reading connection-local state rather than the file header. Test 1 opens a *new* `SqliteConnection` after the helper runs, then queries `PRAGMA journal_mode;` — proves WAL persisted in the file header.

### Windows-specific: `SqliteConnection.ClearAllPools()`

Microsoft.Data.Sqlite pools connections per connection-string. On Windows the file handle stays held even after `Dispose`, so `File.Delete` fails. Added `ClearAllPools()` to the test's `finally` block. Discovered during initial test run — the test failed with `System.IO.IOException: process cannot access the file` until pools were cleared.

## Test results

- `SqlitePragmaTests` — **3/3 passed** (1 file-backed WAL persistence verification, 2 in-memory no-throw cases via `[DataTestMethod]`).
- `EventSourcingIntegrationTests` regression baseline — **6/6 passed** (24s). In-memory SQLite via `WorkflowTestBase`, where the WAL pragma returns `"memory"` no-op.

## Test plan

- [ ] CI green
- [ ] Optional: after merge, observe Api+Web log volume around SQLITE_BUSY (should drop to zero under normal concurrent load).

## Out of scope (per round-2 plan)

- `PRAGMA synchronous=NORMAL` on the query context — separate optimization, perf-driven, requires per-context configuration.
- EF query-side retry policy — WAL + 30s default busy_timeout should eliminate the original SQLITE_BUSY failure mode.
- The `DbConnection` overload of `UseFleansSqlite` — test base owns the connection in that path.
- Postgres counterpart — SQLite-specific issue.

### DIFF
--- a/src/Fleans/Fleans.Persistence.Sqlite/SqliteSchemaInitializer.cs
+++ b/src/Fleans/Fleans.Persistence.Sqlite/SqliteSchemaInitializer.cs
@@ -28,5 +28,39 @@ public static class SqliteSchemaInitializer
         {
             // Tables already created by a concurrent process sharing the same SQLite file.
         }
+
+        // Enable WAL once. Sticky across connections via the database header — subsequent
+        // opens on any thread or process see the file in WAL mode without re-running the
+        // pragma. Idempotent: re-running on an already-WAL file is a no-op.
+        //
+        // PRAGMA returns the resulting mode as a 1-row result set, so we use ExecuteScalar
+        // (not ExecuteNonQuery — some provider versions treat that as undefined behaviour
+        // for PRAGMA).
+        //
+        // busy_timeout intentionally NOT set here: Microsoft.Data.Sqlite already defaults
+        // SqliteCommand.CommandTimeout to 30s and internally calls sqlite3_busy_timeout
+        // before each command. The PRAGMA-level setting would be per-connection and not
+        // worth the maintenance cost given the provider default suffices.
+        var conn = (SqliteConnection)database.GetDbConnection();
+        var wasClosed = conn.State != System.Data.ConnectionState.Open;
+        try
+        {
+            if (wasClosed) conn.Open();
+            using var cmd = conn.CreateCommand();
+            cmd.CommandText = "PRAGMA journal_mode=WAL;";
+            _ = cmd.ExecuteScalar(); // returns "wal" for file-backed, "memory" for in-memory — we don't care which
+        }
+        catch (SqliteException ex)
+        {
+            // Likely an in-memory or read-only DB. WAL doesn't apply to those; safe to
+            // proceed. Surface the message so a misconfigured writable DB (e.g.,
+            // directory permission lost between deploys) isn't silently downgraded to
+            // DELETE-mode journaling.
+            Console.Error.WriteLine($"[Fleans] WARNING: Failed to set journal_mode=WAL: {ex.Message}");
+        }
+        finally
+        {
+            if (wasClosed && conn.State == System.Data.ConnectionState.Open) conn.Close();
+        }
     }
 }
--- a/src/Fleans/Fleans.Persistence.Tests/SqlitePragmaTests.cs
+++ b/src/Fleans/Fleans.Persistence.Tests/SqlitePragmaTests.cs
@@ -1,0 +1,72 @@
+using Fleans.Persistence.Events;
+using Fleans.Persistence.Sqlite;
+using Microsoft.Data.Sqlite;
+using Microsoft.EntityFrameworkCore;
+
+namespace Fleans.Persistence.Tests;
+
+/// <summary>
+/// #656 — verifies <c>SqliteSchemaInitializer.EnsureCreatedIgnoreRaces</c> sets
+/// <c>PRAGMA journal_mode=WAL</c> on file-backed databases and silently no-ops
+/// (without throwing) on in-memory databases where WAL is inapplicable.
+/// </summary>
+[TestClass]
+public class SqlitePragmaTests
+{
+    [TestMethod]
+    public void EnsureCreatedIgnoreRaces_SetsWalJournalMode_OnFileBackedDb()
+    {
+        // Arrange — real temp file. In-memory would no-op the pragma ("memory" mode)
+        // and not exercise WAL persistence in the file header.
+        var dbPath = Path.GetTempFileName();
+        try
+        {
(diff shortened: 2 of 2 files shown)


nightBaker (commented):
## Code Review — fix(#656): enable SQLite WAL journal_mode on schema init

Implementation matches the [round-2 plan](https://github.com/nightBaker/fleans/issues/656#issuecomment-4504839932) approved by [round-2 review](https://github.com/nightBaker/fleans/issues/656#issuecomment-4504878010), with both 🟢 review nudges incorporated and one empirical finding documented.

### What's solid

- ✅ **Edit 1 mechanics**: `PRAGMA journal_mode=WAL` via direct `SqliteCommand` + `ExecuteScalar` (the documented PRAGMA pattern, avoiding the round-1 `ExecuteSqlRaw`-might-no-op concern). Connection-state preservation with `wasClosed` flag + conditional `Close()` in `finally`.
- ✅ **Edit 2 drop justified**: pre-impl `grep CommandTimeout src/Fleans/Fleans.Persistence/` confirmed zero overrides — the 30s default flows through to every command via `SqliteCommand.CommandTimeout`. Documented in commit body + a brief code comment in `SqliteSchemaInitializer.cs`.
- ✅ **Round-2 review 🟢 #1 addressed**: `[DataTestMethod]` covers both `:memory:` and `file::memory:?cache=shared` (the form `WorkflowTestBase` uses).
- ✅ **Round-2 review 🟢 #2 addressed**: pre-impl grep confirmed busy_timeout drop rationale.
- ✅ **Fresh verification connection** in `EnsureCreatedIgnoreRaces_SetsWalJournalMode_OnFileBackedDb` — proves WAL persisted in the file header, addressing round-1's spurious-pass concern.
- ✅ **Observability via `Console.Error.WriteLine`** on PRAGMA failure (operator-visible signal for silently-misconfigured prod DBs).

### Empirical finding documented in this PR

**`SqliteConnection.ClearAllPools()` required in test cleanup on Windows.** Discovered during initial test run: `File.Delete` failed with `process cannot access the file` because Microsoft.Data.Sqlite pools file handles per connection-string and `Dispose` doesn't release them. Comment in code explains; `ClearAllPools()` added to test's `finally` before the file deletes. Reasonable to log as a feedback-memory item if this comes up across multiple cycles.

### CLAUDE.md compliance

- ✅ Records / immutable DTOs — N/A here (test-only and bootstrap-init code).
- ✅ Dynamic state via `ExpandoObject` — N/A.
- ✅ `[LoggerMessage]` source generators — N/A; `SqliteSchemaInitializer` is a static low-level bootstrap helper with no `ILogger` dependency. Using `Console.Error.WriteLine` matches the established pattern in `Fleans.Persistence.Sqlite` (verified inline). The CLAUDE.md rule is about application code; bootstrap helpers are exempt.
- ✅ Tests follow AAA pattern with clear Arrange/Act/Assert sections.
- ✅ No `WorkflowInstance` event-sourcing concerns (no production code touched outside `Fleans.Persistence.Sqlite`).
- ✅ No streaming / grain-state-mutation invariants applicable.

### Test results (per PR description)

- `SqlitePragmaTests` — **3/3 passed** (1 file-backed WAL persistence verification, 2 in-memory no-throw cases).
- `EventSourcingIntegrationTests` regression baseline — **6/6 passed** (24s).

### 🟢 Minor (non-blocking)

1. **The Microsoft.Data.Sqlite pool comment in the test could become a feedback-memory entry.** Saving "Windows SQLite tests: always `SqliteConnection.ClearAllPools()` before `File.Delete` in `finally`" would prevent future cycles from re-debugging the `process cannot access the file` error. Implementer's call; out of scope for this PR.

2. **Edit 2's drop rationale is currently encoded only in commit message + a brief code comment.** Worth a one-line note in `docs/conventions/persistence.md` next time that file is touched, so a future reader understands why the WAL pragma is set explicitly but `busy_timeout` is not. Not blocking — the commit history is sufficient.

### Verdict

No 🔴, no 🟡. CI is PENDING; once it flips green, the worker will auto-promote to Testing.

OK to promote.

PR #671, opened 2026-05-21
feat(#659): defend CustomTaskCatalogGrain against silent self-wipe

### LINKED ISSUE #659: CustomTaskCatalogGrain activation-read failure can wipe persisted catalog with {}
## Problem
`CustomTaskCatalogGrain.OnActivateAsync` reads persisted state from `IPersistentState`. If `ReadStateAsync` fails (transient DB error, deserialization), the grain may come up with empty in-memory state, and the **next write overwrites the real persisted catalog with `{}`**. This is a self-destructing failure mode: one transient read error → permanent state loss.

## Evidence
- `src/Fleans/Fleans.Application/Grains/CustomTaskCatalogGrain.cs:44-56` — activation read path.
- `src/Fleans/Fleans.Application/Grains/CustomTaskCatalogGrain.cs:70, 143` — write paths trust the in-memory state.

## Impact
- One bad read during activation → catalog wiped on next write.
- The catalog drives custom-task plugin discovery; wiping it breaks every plugin-backed workflow until re-registered manually.

## Suggested fix
1. **Fail activation on read failure.** Re-throw from `OnActivateAsync` so Orleans retries activation later. Do **not** silently start with empty state when a read failed.
2. Add a unit test: simulate `ReadStateAsync` throwing on activation, assert the grain refuses to come up and the persisted state is untouched.
3. Audit other grains for the same pattern: `rg -n "OnActivateAsync" src/Fleans/Fleans.Application/Grains/` and inspect each for swallowed read failures.

## Source

nightBaker:
Closes #659.

## Summary

`CustomTaskCatalogGrain.ReconcileWithMembership` (`Fleans.Application/CustomTasks/CustomTaskCatalogGrain.cs:128-160`) used to strip every catalog entry — and persist the empty state — when `IManagementGrain.GetDetailedHosts(onlyActive: false)` returned a successful but anomalous empty list. That signature is overwhelmingly a transient Redis/membership-table blip, not a real "every silo left the cluster" event, so the persisted catalog was being silently wiped on a non-fatal hiccup.

This PR adds 5 layered defenses against that wipe path, per round-2 design.

## Layers (round-2 design ↔ this PR)

- **A — `ReconcileGuard.IsAnomalousEmptyAliveSilos` (pure helper)**: extracted as an `internal static` class above the grain. `ReconcileWithMembership` now calls it between `aliveSilos` computation and `RemoveWhere`. If the guard returns true (alive=0 ∧ entries>0), the reconcile tick logs at Warning level (EventId 9323, `LogReconcileSkippedAnomalousEmptyAliveSilos`) and returns without writing.
- **B — `EfCoreCustomTaskCatalogGrainStorage.ReadStateAsync` symmetric assignment**: removes the asymmetric `if (rows.Count > 0) { … }` branch; state is always assigned, and ETag/`RecordExists` are explicit non-loaded markers when the table is empty. No behavioural change on the happy path.
- **C — Read-failure propagation test**: `ReadStateAsync_DbContextFactoryThrows_PropagatesException` in `EfCoreCustomTaskCatalogGrainStorageTests` proves a throwing `IDbContextFactory` (`ThrowingDbContextFactory` inlined as `private sealed class` in the test file) propagates its exception and leaves `RecordExists=false` / `ETag=null`. Provider-agnostic — no `[DataRow]` because the exception happens before any DB interaction.
- **D — `ReconcileGuardTests`**: 4 pure-function unit tests covering the (alive × entries) {0, >0} matrix. No TestCluster, no `IManagementGrain` mocking.
- **E — Spot-check on flagged peers**: checked `ConditionalStartEventRegistryGrain` and `TimerStartEventSchedulerGrain` inline. Neither reconciles against external membership state — `ConditionalStartEventRegistryGrain` mutates only on explicit caller-driven `Register` / `Unregister[All]`; `TimerStartEventSchedulerGrain` writes only on caller-driven `Activate` / `FireTimerStartEvent`. CustomTaskCatalogGrain is the only grain with the wipe-prone shape.

## Visibility note

`ReconcileGuard` is `internal static` because it's a defensive helper, not a public contract. To let Layer D target it, added `<InternalsVisibleTo Include=\"Fleans.Application.Tests\" />` to `Fleans.Application.csproj`. First InternalsVisibleTo in the repo — preferred over making the helper public because it's a pure private invariant of the grain's reconcile path.

## How to test

```bash
cd src/Fleans
dotnet build
dotnet test Fleans.Application.Tests/Fleans.Application.Tests.csproj --filter \"FullyQualifiedName~ReconcileGuard|FullyQualifiedName~CustomTaskCatalog\"
dotnet test Fleans.Persistence.Tests/Fleans.Persistence.Tests.csproj --filter \"FullyQualifiedName~CustomTaskCatalog\"
```

Local results: Application.Tests 338/338 passed (6 skipped, 0 failed), Persistence.Tests 148/148 passed (124 PG rows skipped without `FLEANS_PG_TESTS=1`, 0 failed).

## Out of scope (filed elsewhere if confirmed)

- Partial-disjoint anomaly: `aliveSilos.Count > 0` but disjoint from registered silos still wipes all. Requires a different signal (threshold-based removal count or external "minimum cluster size"). Acknowledged in round-1 review as 🟢 minor; not blocking this PR.

### DIFF
--- a/src/Fleans/Fleans.Application.Tests/CustomTasks/ReconcileGuardTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/CustomTasks/ReconcileGuardTests.cs
@@ -1,0 +1,31 @@
+using Fleans.Application.CustomTasks;
+
+namespace Fleans.Application.Tests.CustomTasks;
+
+[TestClass]
+public class ReconcileGuardTests
+{
+    [TestMethod]
+    public void IsAnomalousEmptyAliveSilos_ZeroSilosWithEntries_ReturnsTrue()
+    {
+        Assert.IsTrue(ReconcileGuard.IsAnomalousEmptyAliveSilos(aliveSilosCount: 0, currentEntryCount: 3));
+    }
+
+    [TestMethod]
+    public void IsAnomalousEmptyAliveSilos_ZeroSilosWithZeroEntries_ReturnsFalse()
+    {
+        Assert.IsFalse(ReconcileGuard.IsAnomalousEmptyAliveSilos(aliveSilosCount: 0, currentEntryCount: 0));
+    }
+
+    [TestMethod]
+    public void IsAnomalousEmptyAliveSilos_SomeSilosWithEntries_ReturnsFalse()
+    {
+        Assert.IsFalse(ReconcileGuard.IsAnomalousEmptyAliveSilos(aliveSilosCount: 2, currentEntryCount: 3));
+    }
+
+    [TestMethod]
+    public void IsAnomalousEmptyAliveSilos_SomeSilosWithZeroEntries_ReturnsFalse()
+    {
+        Assert.IsFalse(ReconcileGuard.IsAnomalousEmptyAliveSilos(aliveSilosCount: 2, currentEntryCount: 0));
+    }
+}
--- a/src/Fleans/Fleans.Application/CustomTasks/CustomTaskCatalogGrain.cs
+++ b/src/Fleans/Fleans.Application/CustomTasks/CustomTaskCatalogGrain.cs
@@ -8,6 +8,24 @@ using Orleans.Runtime;
 namespace Fleans.Application.CustomTasks;
 
 /// <summary>
+/// Pure logic for the catalog reconcile sanity guard. Extracted so unit tests
+/// can target the guard's decision directly without standing up a TestCluster
+/// or mocking <see cref="IManagementGrain"/>. See #659.
+/// </summary>
+internal static class ReconcileGuard
+{
+    /// <summary>
+    /// Detects the "membership directory anomaly" case where the catalog has
+    /// entries but the management grain reports zero alive silos — almost
+    /// certainly a transient Redis/membership-table blip, not a real "every
+    /// silo left the cluster" event. The reconciler must skip writes in this
+    /// case to avoid silently wiping the persisted catalog.
+    /// </summary>
+    public static bool IsAnomalousEmptyAliveSilos(int aliveSilosCount, int currentEntryCount)
+        => aliveSilosCount == 0 && currentEntryCount > 0;
+}
+
+/// <summary>
 /// Catalog of custom-task plugin registrations announced by Worker silos.
 /// State is persisted via <see cref="IPersistentState{T}"/> backed by an EF Core
 /// grain-storage provider (sub-issue A2 of #357 / PR #434), so the catalog survives
@@ -137,6 +155,12 @@ public sealed partial class CustomTaskCatalogGrain : Grain, ICustomTaskCatalogGr
                 .Where(n => !string.IsNullOrEmpty(n))
                 .ToHashSet(StringComparer.Ordinal);
 
+            if (ReconcileGuard.IsAnomalousEmptyAliveSilos(aliveSilos.Count, _state.State.Entries.Count))
+            {
+                LogReconcileSkippedAnomalousEmptyAliveSilos(_state.State.Entries.Count);
+                return;
+            }
+
             var removed = _state.State.RemoveWhere(e => !aliveSilos.Contains(e.SiloName));
             if (removed > 0)
             {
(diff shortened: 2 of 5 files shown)


nightBaker (commented):
## Code Review — feat(#659): defend CustomTaskCatalogGrain against silent self-wipe

Reviewed the 5-file diff against CLAUDE.md and the round-2 design. Each layer matches the plan; correctness of the guard and the storage refactor verified inline. No blocking issues.

### 🟢 Minor

1. **Layer B has an implicit-Orleans-default dependency worth noting.** The old `ReadStateAsync` left `grainState.State` unassigned on empty DB; the new code now explicitly sets it to `new CustomTaskCatalogState { Entries = rows }`. This matches what Orleans does on its own (default-construct the state type before calling `ReadStateAsync`), so the change is a no-op refactor in practice — the existing `CustomTaskCatalogGrain.OnActivateAsync → ReconcileWithMembership → _state.State.Entries.Count` path only works today because of that Orleans default. The symmetric assignment makes the storage less dependent on that implicit behaviour. Worth documenting in the `ReadStateAsync` xmldoc if anyone touches it again.

2. **Empty-DB read path isn't asserted post-Layer-B.** The existing `ReadStateAsync_EmptyDatabase_ReturnsNoRecord` test still asserts `IsFalse(RecordExists)` + `HasCount(0, State.Entries)` — both pass under old and new code. It doesn't verify that `State` is non-null after Layer B's symmetric assignment. Not a blocker because `OnActivateAsync` exercises that path on every cold start, but a one-line `Assert.IsNotNull(read.State)` would future-proof the empty-DB contract.

3. **`ReconcileGuard` internal + first `InternalsVisibleTo` in the repo.** The PR description flags this as a new convention. I'm fine with it — making a defensive private invariant `public` just for test access would be worse — but if the project drifts toward more `internal` helpers under test, it's worth a one-line note in `docs/conventions/` once a second case lands. Don't preempt it for this PR.

### Layer-by-layer verification

- **A (`ReconcileGuard.IsAnomalousEmptyAliveSilos` + skip-with-log)** — Pure function with the right predicate (`alive==0 && entries>0`). Insertion point in `ReconcileWithMembership` is between `aliveSilos` computation and `RemoveWhere`, which is exactly where the design said. Warning-level log with EventId 9323 (next-free slot per the design's audit). The wipe path is genuinely blocked: `return;` skips both `RemoveWhere` and `WriteStateAsync`.
- **B (symmetric `ReadStateAsync` assignment)** — Clean. See 🟢 #1 above.
- **C (read-failure propagation test)** — Inline `ThrowingDbContextFactory` is correct; both factory methods throw, so the storage's `await using var db = await _dbContextFactory.CreateDbContextAsync()` triggers the exception before any DB call. Assertions correctly cover both behaviours (exception propagates AND state markers stay non-loaded).
- **D (`ReconcileGuardTests`)** — 4 tests cover the (alive × entries) {0, >0} matrix. No TestCluster, no `IManagementGrain` mocking — exactly the right scope-cost tradeoff per the round-2 design's rationale.
- **E (audit of peer grains)** — `ConditionalStartEventRegistryGrain` and `TimerStartEventSchedulerGrain` confirmed non-reconcile patterns — caller-driven mutations only, no `IManagementGrain.GetDetailedHosts → RemoveWhere` shape. CustomTaskCatalogGrain remains the only grain exposed to this class of bug.

### CLAUDE.md compliance

- ✅ `[LoggerMessage]` source generator — EventId 9323 declared via partial method on the partial class. No `ILogger.Log*()` extension calls.
- ✅ Domain-state invariants — no state mutations bypassed; `_state.State.RemoveWhere` is still the only mutation entry point, now correctly skipped under the anomaly.
- ✅ Tests follow the AAA pattern; activity-test rule N/A (this is a grain-storage + reconcile guard, not an activity).
- ✅ Documentation scope split — no website docs touched, no source-tree pins. Bug fix, not a new feature, so no manual test plan required per the conventions in `tests/manual/README.md`.

PR #663, opened 2026-05-20
feat(#650): persistent reminder service (Redis) for BPMN timer durability

nightBaker:
## Summary

Closes #650. Replaces `UseInMemoryReminderService()` (BPMN timers silently dropped on every silo restart) with the Redis-backed Orleans reminder service. Per round-3 plan ([approved here](https://github.com/nightBaker/fleans/issues/650#issuecomment-4502335116)).

### What ships

- **New package**: `Microsoft.Orleans.Reminders.Redis` 10.0.1 added to `Fleans.ServiceDefaults`.
- **New helper**: `Fleans.ServiceDefaults/Reminders/FleansRemindersExtensions.cs` — `AddFleansReminders(IConfiguration)`. Throws `InvalidOperationException` at startup if `orleans-redis` connection string is missing (per "Registration-vs-cleanup error asymmetry" — silent fallback to in-memory would re-create the exact bug this PR fixes).
- **New options shim**: `FleansReminderOptions.cs` — pins the canonical connection name.
- **3 call-site swaps**:
  - `Fleans.Api/Program.cs`: `UseInMemoryReminderService()` → `AddFleansReminders(builder.Configuration)`.
  - `Fleans.WorkerHost/Program.cs`: same.
  - `Fleans.Aspire/Program.cs`: `.WithMemoryReminders()` deleted; AppHost is orchestration-only, the silo configures reminders itself.
- **Docs**: new `docs/conventions/reminders.md` covers provider choice, fail-fast semantics, cluster invariants, migration note, and the `TimerCallbackGrain` 1-minute-granularity clarifier. CLAUDE.md gains one-line pointer.
- **Manual test plan #65**: `tests/manual/65-persistent-reminders/` with `timer-restart.bpmn` (Start → TimerICE PT5M → End). Step A single-silo restart, Step B multi-silo Core-only restart, plus fail-fast regression with `orleans-redis` removed.

### Why Redis (not AdoNet)

Redis is already required for clustering + streaming — no new infrastructure. `Microsoft.Orleans.Reminders.AdoNet` does not support SQLite, which would force every SQLite quick-start user to install Postgres for timer durability and break the SQLite single-binary story. Reminders ride the clustering Redis, decoupled from the application persistence pivot (`FLEANS_PERSISTENCE_PROVIDER={Sqlite,Postgres}`).

### Implementation notes

[...]

PR #672, opened 2026-05-21
feat(#669): add Postgres provider option for Orleans reminders

### LINKED ISSUE #669: Add Postgres provider option for Orleans reminders (opt-in alongside default Redis)
## Problem

PR #663 (closes #650) shipped Redis-backed Orleans reminders as the only persistent option, with this rationale in the PR body:

> "Microsoft.Orleans.Reminders.AdoNet does not support SQLite, which would force every SQLite quick-start user to install Postgres for timer durability and break the SQLite single-binary story. Reminders ride the clustering Redis, decoupled from the application persistence pivot."

That rationale stands for the SQLite default — but doesn't preclude an **opt-in Postgres-backed reminders** for operators already running Postgres-application who want uniform persistence semantics (backup, replication, point-in-time recovery) for their timer state too. `Microsoft.Orleans.Reminders.AdoNet` supports Postgres via the `Npgsql` invariant cleanly.

## Goal

Add `Postgres` as a second value for a new `FLEANS_REMINDERS_PROVIDER` env var (default `Redis` — back-compat with what #663 just shipped). Operators who already provision Postgres for application persistence (`FLEANS_PERSISTENCE_PROVIDER=Postgres`) can opt their reminders into the same store.

## Out of scope

- **SQLite reminders** — `Microsoft.Orleans.Reminders.AdoNet` does not bundle a SQLite invariant. SQLite quick-start users continue to get Redis reminders.
- **Removing Redis from the stack** — Redis is still required for clustering + streaming (both default to Redis). Postgres reminders alone does NOT free operators from Redis. A "no Redis at all" architecture is a separate, much larger redesign.

## Suggested fix (shape)

1. New env var `FLEANS_REMINDERS_PROVIDER` with values `Redis` (default) and `Postgres`. Mirror the validation pattern in `FleanStreamingExtensions.cs:26` — invalid values throw at startup.
2. Add `Microsoft.Orleans.Reminders.AdoNet` package to `Fleans.ServiceDefaults`.
3. Extend `FleansRemindersExtensions.AddFleansReminders(IConfiguration)`:
   - Switch on the provider value.
   - Redis branch: existing behavior (current `UseRedisReminderService` call, fail-fast on missing `orleans-redis` connection string).
   - Postgres branch: validate `ConnectionStrings:fleans` is a Postgres connection string (or add a new dedicated `ConnectionStrings:reminders` if mixing apps + reminders on one DB is undesirable), then call `UseAdoNetReminderService(opt => { opt.Invariant = "Npgsql"; opt.ConnectionString = ...; })`. Fail-fast if SQLite is configured alongside Postgres reminders (combination is unsupported).
4. Document in `docs/conventions/reminders.md` — provider matrix, what `OrleansReminderTable_{ServiceId}` table is, and Orleans's auto-create-on-first-run behavior (or whether operators must run an `init.sql`).

## Test plan

- Unit test for `AddFleansReminders` provider-switch shape (the wiring unit test already deferred in #663's follow-up list — this issue subsumes that work).
- Extend manual test plan `tests/manual/65-persistent-reminders/test-plan.md` with a "Postgres reminders" step: deploy `timer-restart.bpmn` against `FLEANS_REMINDERS_PROVIDER=Postgres`, restart `fleans-core`, confirm timer fires post-restart.
- E2E suite: at least one parameterized run against Postgres-reminders (CI matrix-style, similar to how `FLEANS_PG_TESTS=1` exercises the pg-persistence row today).

## Impact

- **Operator benefit**: uniform Postgres backup/replication for application data + reminders. Narrow but real for Postgres-only-application operators.
- **New config dimension**: matrix grows to `Persistence × Streaming × Reminders` = SQLite/Postgres × Redis/Kafka/Memory/AzureQueue × Redis/Postgres. Need at least one matrix-row in CI to keep the Postgres-reminders path green.

## Related

- Closes a follow-up gap from #650 / #663.
- Adjacent to #459 (Swap ProjectReference → PackageReference) and #458 (NuGet signing) only in that all three touch deployment surface — not blocking either.

## Priority

P2 — opt-in enhancement, not a correctness bug. Existing Redis-reminders path is the supported default.

nightBaker:
Closes #669.

## Summary

Adds opt-in **Postgres-backed Orleans reminders** alongside the default Redis reminders shipped in #663. Operators on `Persistence:Provider=Postgres` can now uniform their app + reminder durability on Postgres by setting `Fleans:Reminders:Provider=Postgres`.

Per the round-3 design — 3 analysis rounds caught real flaws (broken sentinel under EF Core's `ExecuteSqlRawAsync` SELECT semantics, manifest-resource naming defaults, Apache-2.0 attribution specifics, vendored-script dependency on `OrleansQuery` from `PostgreSQL-Main.sql`).

## Provider matrix

| `Persistence:Provider` | `Fleans:Reminders:Provider` | Outcome |
|---|---|---|
| `Sqlite` (default) | `Redis` (default) | Supported. Single-binary story preserved. |
| `Sqlite` | `Postgres` | **Unsupported — silo refuses to start.** |
| `Postgres` | `Redis` (default) | Supported. Reminders ride the clustering Redis. |
| `Postgres` | `Postgres` | Supported. Reminders share `ConnectionStrings:fleans` with app persistence. |

## What ships

- New config key `Fleans:Reminders:Provider` (matched case-insensitively; default `Redis` — back-compat with #663). Bound to a new `FleansPersistenceOptions.RemindersProvider` field at registration time so `EnsureDatabaseSchemaAsync` reads it via `IOptions` without re-accessing `IConfiguration` after `Build()`.
- `FleansRemindersExtensions.AddFleansReminders` becomes provider-switching with three fail-fast guards: (a) Postgres+SQLite-app combo, (b) missing connection string for the chosen provider, (c) unknown provider value.
- `Microsoft.Orleans.Reminders.AdoNet 10.0.1` added to `Fleans.ServiceDefaults`, **pinned** to match the `Microsoft.Orleans.*` sibling versions.
- `PostgreSQL-Main.sql` + `PostgreSQL-Reminders.sql` vendored unmodified from the upstream Orleans `v10.0.1` tag into `src/Fleans/Fleans.ServiceDefaults/Resources/` (Apache-2.0; per-file header + `LICENSE-orleans-vendor` at the repo root).
- `EnsureDatabaseSchemaAsync` runs both scripts under the existing advisory lock when Postgres reminders are enabled. Idempotent via the `OrleansQuery` existence sentinel checked via raw ADO.NET `ExecuteScalarAsync` (EF Core's `ExecuteSqlRawAsync` returns -1 on SELECT in Npgsql so it cannot be used as a sentinel).

## Tests

- `Fleans.ServiceDefaults.Tests/FleansRemindersExtensionsTests.cs` — 6 tests covering the provider-switch + fail-fast matrix via the testable `ResolveRemindersConfiguration()` pure function (extracted so unit tests don't need a TestCluster).
- `Fleans.ServiceDefaults.Tests/EmbeddedSchemaResourceTests.cs` — 4 tests (via `[DataRow]`) verifying (a) both vendored .sql files are embedded under the expected manifest names and (b) carry the version-pin comment as a drift guard against silent SDK bumps that forget to re-vendor.
- All `Fleans.ServiceDefaults.Tests` (15/15) and `Fleans.Persistence.Tests` (147/147 — 124 PG rows skipped without `FLEANS_PG_TESTS=1`) pass locally.

## Docs + manual

- `docs/conventions/reminders.md` gets a provider matrix and a "Postgres reminders schema" subsection covering the vendored-script approach and the idempotent sentinel.
- `tests/manual/65-persistent-reminders/test-plan.md` gets **Step C** (Postgres reminders single-silo restart with table-creation verification queries) and **Step D** (mixed-storage fail-fast regression).

## Out of scope (filed as follow-ups if a real operator asks)

- Dedicated `ConnectionStrings:reminders` override (separate Postgres instance from app persistence).
- SQLite reminders provider (`Microsoft.Orleans.Reminders.AdoNet` doesn't bundle a SQLite invariant; upstream SDK work).
- Helm chart `reminders.provider` knob (separate issue against the chart repo).

## Re-vendoring runbook

When `Microsoft.Orleans.Reminders.AdoNet` is bumped, the two `.sql` files in `Resources/` must be re-fetched from the matching upstream tag and their version-pin comment updated. The `EmbeddedResource_ContainsVersionPin` test will fail immediately if a bump forgets the re-vendor step.

### DIFF
--- a/LICENSE-orleans-vendor
+++ b/LICENSE-orleans-vendor
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
+      cross-claim or counterclaim in a lawsuit) alleging that the Work
+      or a Contribution incorporated within the Work constitutes direct
+      or contributory patent infringement, then any patent licenses
+      granted to You under this License for that Work shall terminate
+      as of the date such litigation is filed.
+
+   4. Redistribution. You may reproduce and distribute copies of the
+      Work or Derivative Works thereof in any medium, with or without
+      modifications, and in Source or Object form, provided that You
+      meet the following conditions:
+
+      (a) You must give any other recipients of the Work or
+          Derivative Works a copy of this License; and
+
+      (b) You must cause any modified files to carry prominent notices
+          stating that You changed the files; and
+
+      (c) You must retain, in the Source form of any Derivative Works
+          that You distribute, all copyright, patent, trademark, and
+          attribution notices from the Source form of the Work,
+          excluding those notices that do not pertain to any part of
+          the Derivative Works; and
+
+      (d) If the Work includes a "NOTICE" text file as part of its
+          distribution, then any Derivative Works that You distribute must
+          include a readable copy of the attribution notices contained
+          within such NOTICE file, excluding those notices that do not
+          pertain to any part of the Derivative Works, in at least one
+          of the following places: within a NOTICE text file distributed
+          as part of the Derivative Works; within the Source form or
+          documentation, if provided along with the Derivative Works; or,
+          within a display generated by the Derivative Works, if and
+          wherever such third-party notices normally appear. The contents
+          of the NOTICE file are for informational purposes only and
+          do not modify the License. You may add Your own attribution
+          notices within Derivative Works that You distribute, alongside
+          or as an addendum to the NOTICE text from the Work, provided
+          that such additional attribution notices cannot be construed
+          as modifying the License.
+
+      You may add Your own copyright statement to Your modifications and
+      may provide additional or different license terms and conditions
+      for use, reproduction, or distribution of Your modifications, or
+      for any such Derivative Works as a whole, provided Your use,
+      reproduction, and distribution of the Work otherwise complies with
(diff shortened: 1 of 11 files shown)


nightBaker (commented):
## Code Review — feat(#669): Postgres provider option for Orleans reminders

Reviewed the 11-file PR against CLAUDE.md, the round-3 design, and the round-3 review notes. Production code matches the plan; tests cover the provider-switch + fail-fast surface (6 tests) and the embedded-resource layout + version-pin drift guard (4 tests). 10/10 new tests pass, full ServiceDefaults.Tests (15/15) and Persistence.Tests (147/147) green.

**No 🔴 blocking. One 🟡 important + one 🟢 minor.**

### 🟡 Important

**1. Schema-init is non-atomic — partial-init can permanently brick the silo.**

`FleansPersistenceExtensions.cs:108-112` runs both vendored scripts sequentially without an explicit transaction:

```csharp
if (!alreadyInitialised)
{
    await ExecuteEmbeddedScriptAsync(db, "Fleans.ServiceDefaults.Resources.PostgreSQL_Main.sql");
    await ExecuteEmbeddedScriptAsync(db, "Fleans.ServiceDefaults.Resources.PostgreSQL_Reminders.sql");
}
```

`ExecuteSqlRawAsync` executes each script's full multi-statement string outside an explicit transaction. If a statement mid-script fails (transient connection blip, statement-timeout, OOM during `INSERT INTO OrleansQuery`), prior statements remain committed.

Failure cascade:

1. First start: `CREATE TABLE OrleansQuery` succeeds, then a subsequent `INSERT INTO OrleansQuery (...)` errors. Exception propagates → advisory lock released → silo refuses to start. (So far OK — fail-fast asymmetry holds.)
2. Operator restarts the silo. The sentinel `SELECT EXISTS (SELECT 1 FROM pg_class WHERE relname = 'orleansquery')` returns true → schema-init is skipped → silo starts cleanly.
3. AdoNet reminder writes fail forever because the query templates were never fully populated. Operator can't recover without manually dropping the partial schema.

Fix is a one-block change — wrap both `ExecuteEmbeddedScriptAsync` calls in an explicit transaction so partial-init rolls back:

```csharp
if (!alreadyInitialised)
{
    await using var initTx = await db.Database.BeginTransactionAsync();
    await ExecuteEmbeddedScriptAsync(db, "Fleans.ServiceDefaults.Resources.PostgreSQL_Main.sql");
    await ExecuteEmbeddedScriptAsync(db, "Fleans.ServiceDefaults.Resources.PostgreSQL_Reminders.sql");
    await initTx.CommitAsync();
}
```

Postgres supports `CREATE FUNCTION` inside a transaction (the upstream Main.sql header notes "requires Postgres 9.5 or perhaps higher"), so no DDL-restriction issue. The fix is small enough I'd accept it in a follow-up PR if it's easier to handle independently, but the current code's "happy path works" claim is conditional on every multi-statement `ExecuteSqlRawAsync` running without interruption.

Not blocking promotion because the failure mode requires a mid-script error AND a subsequent restart AND an operator who doesn't notice the silo refused to start on first run — narrow window. But worth fixing soon.

### 🟢 Minor

**1. `[DataTestMethod]` is deprecated in MSTest 4.**

`EmbeddedSchemaResourceTests.cs:13, :24` use `[DataTestMethod]` which triggers an `MSTEST0044` warning at build time ("DataTestMethod" is deprecated; use "TestMethod"). MSTest 4's `[TestMethod]` correctly composes with `[DataRow]`. Replace `[DataTestMethod]` with `[TestMethod]` (two spots). Trivial. The warning is visible in the build output for this PR.

### Verifications

- ✅ CLAUDE.md compliance: no website docs touched (engine-side change only); EventId reservation N/A (no new `[LoggerMessage]` in this PR); load-bearing invariants unaffected.
- ✅ Provider-switch logic in `ResolveRemindersConfiguration` (`Reminders/FleansRemindersExtensions.cs:53-91`) — clean, all 4 throw paths exercised by 6 unit tests (5 + 1 happy-path Postgres).
- ✅ `RemindersProvider` binding in `FleansPersistenceExtensions.cs:30-35` — explicit non-empty check before overriding the `Redis` default, matches the existing `Provider` binding pattern.
- ✅ Sentinel uses raw ADO.NET `ExecuteScalarAsync` (correctly avoiding the `ExecuteSqlRawAsync` -1-on-SELECT pitfall from round-2/3 review).
- ✅ Vendored .sql files carry the version-pin comment (`Microsoft.Orleans.Reminders.AdoNet v10.0.1`); drift test asserts presence.
- ✅ `LICENSE-orleans-vendor` carries the full Apache-2.0 text (202 lines); per-file headers correctly cite source URL + license + re-vendor instruction.
- ✅ Fail-fast on SQLite-app + Postgres-reminders (D4); fail-fast on missing connection string (D2); error messages mention the relevant config keys for operator debuggability.
- ✅ Manual test plan extended with Step C (Postgres reminders happy path with table-creation verification queries — corrected to `OrleansRemindersTable` per round-2 fix) and Step D (mixed-storage fail-fast regression).
- ✅ Docs matrix in `docs/conventions/reminders.md` is correct; the new "Postgres reminders schema" subsection covers the vendored-script approach and the idempotent sentinel.

### Decision

No blocking issues. OK to promote. The 🟡 partial-init concern should land as a small follow-up PR — recommend tracking it as a low-priority follow-up issue rather than blocking #669's promotion.

PR #694, opened 2026-05-27
fix(#611): replace unbounded GetAllProcessDefinitions in Editor with targeted max-version query

### LINKED ISSUE #611: [audit:scalability] Editor.razor calls non-paged GetAllProcessDefinitions() with no row limit
## Severity

minor

## Finding

`src/Fleans/Fleans.Web/Components/Pages/Editor.razor:435` calls the no-arg overload of `GetAllProcessDefinitions()`:

```csharp
var definitions = await QueryService.GetAllProcessDefinitions();
```

The corresponding `WorkflowQueryService.GetAllProcessDefinitions()` method (`src/Fleans/Fleans.Persistence/WorkflowQueryService.cs`) loads every process definition row with no limit:

```csharp
var definitions = await db.ProcessDefinitions
    .OrderBy(d => d.ProcessDefinitionKey)
    .ThenBy(d => d.Version)
    .ToListAsync();   // ← no Take()
```

This is called when the user opens the BPMN editor's deploy dialog. In a large deployment (hundreds or thousands of deployed BPMN definitions — common in integration or multi-tenant scenarios), this will:
- Stall the UI while materialising the full table, including `BpmnXml` columns which can be several hundred KB each.
- Hold a large object in silo heap for the lifetime of the Blazor circuit.

A paged overload already exists (`GetAllProcessDefinitions(PageRequest page)`) and is used by the REST API and MCP.

## Suggested remediation

Option A (preferred): switch `Editor.razor` to use the paged overload with a reasonable page size (e.g. 200) and add a search/filter input so users can locate a definition without loading all of them.

Option B (quick): add a hard cap to the non-paged overload:

```csharp
.Take(500)
.ToListAsync();
```

and log a warning when the result count equals the cap, prompting the caller to use the paged overload. This is a safety valve, not a fix.

Option A is preferred because the deploy dialog benefits from search-as-you-type and avoids a hard magic-number cap that will eventually be wrong.

nightBaker:
## Summary

Closes #611. The deploy dialog in `Editor.razor` was calling the no-arg `GetAllProcessDefinitions()` and materialising every row (including `BpmnXml`, several hundred KB each) just to compute `max(version) + 1` for a single `processDefinitionKey`. Replaces that with a new `GetMaxVersionByKeyAsync(string)` projection — single round-trip, no `BpmnXml` materialisation.

Implements **Option C** from the approved Design & Plan v3 (https://github.com/nightBaker/fleans/issues/611#issuecomment-4556606901 + v3 reviewer approval).

## Changes

- **`IWorkflowQueryService`**: add `GetMaxVersionByKeyAsync(string)`, remove the no-arg `GetAllProcessDefinitions()`. The paged overload remains.
- **`WorkflowQueryService`**: implement via `Where(...).Select(d => (int?)d.Version).MaxAsync() ?? 0` — projection pushdown skips the `BpmnXml` column entirely.
- **`Editor.razor`**: collapse the 7-line `GetAll → Where → OrderByDescending → FirstOrDefault → +1` chain to a single `GetMaxVersionByKeyAsync` call + `deployNextVersion = $"v{maxVersion + 1}"`.
- **`StubWorkflowQueryService`** (test mock): drop the no-arg stub, add the new method's stub.
- **`WorkflowQueryServiceTests`**: rewrite the two tests that called the removed overload as `GetAllProcessDefinitions_Paged_…` variants using the paged overload with an explicit `Sorts: "processDefinitionKey,version"` to preserve ordering coverage. Add 4 new tests on `GetMaxVersionByKeyAsync` (empty / single / multi / unknown-key).

## Why the test rewrite needs explicit `Sorts`

The paged overload uses Sieve, which does not sort by default. Without `Sorts`, the rewrite read the rows in insertion order and the ordering assertion failed. Added `Sorts: "processDefinitionKey,version"` to make the paged test verify the same ordering shape the removed no-arg method guaranteed.

## API-break note

`IWorkflowQueryService.GetAllProcessDefinitions()` (no-arg) is removed. It is **not** in the published NuGet stack — the interface lives in `Fleans.Application` (solution-internal). Per the issue's maintainer answer, the only production callsite was `Editor.razor:435` (now migrated) and the only test callsites were the 3 sites updated in this PR. `grep -rn "GetAllProcessDefinitions()" src/` returns zero hits after this change.

## Test plan

- [x] `dotnet build` — 0 errors
- [x] `dotnet test Fleans.Persistence.Tests/...` — 154 passed (SQLite), 128 skipped (Postgres, no container), 0 failed
- [x] `dotnet test Fleans.Application.Tests/... --filter "FullyQualifiedName~ProcessDefinitionGrain"` — 4/4 passed
- [ ] Manual smoke matrix (per Design & Plan v3 step 8) — run after deploy:
  - Never-deployed key → expect `deployNextVersion = "v1"`
  - 3 existing deployments of key X → expect `deployNextVersion = "v4"`

### DIFF
--- a/src/Fleans/Fleans.Application.Tests/ProcessDefinitionGrainTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/ProcessDefinitionGrainTests.cs
@@ -167,7 +167,7 @@ namespace Fleans.Application.Tests
         private class StubWorkflowQueryService : IWorkflowQueryService
         {
             public Task<InstanceStateSnapshot?> GetStateSnapshot(Guid workflowInstanceId) => Task.FromResult<InstanceStateSnapshot?>(null);
-            public Task<IReadOnlyList<ProcessDefinitionSummary>> GetAllProcessDefinitions() => Task.FromResult<IReadOnlyList<ProcessDefinitionSummary>>([]);
+            public Task<int> GetMaxVersionByKeyAsync(string processDefinitionKey) => Task.FromResult(0);
             public Task<PagedResult<ProcessDefinitionSummary>> GetAllProcessDefinitions(PageRequest page) => Task.FromResult(new PagedResult<ProcessDefinitionSummary>([], 0, page.Page, page.PageSize));
             public Task<PagedResult<ProcessDefinitionGroup>> GetProcessDefinitionGroups(PageRequest page) => Task.FromResult(new PagedResult<ProcessDefinitionGroup>([], 0, page.Page, page.PageSize));
             public Task<PagedResult<WorkflowInstanceInfo>> GetInstancesByKey(string processDefinitionKey, PageRequest page) => Task.FromResult(new PagedResult<WorkflowInstanceInfo>([], 0, page.Page, page.PageSize));
--- a/src/Fleans/Fleans.Application/IWorkflowQueryService.cs
+++ b/src/Fleans/Fleans.Application/IWorkflowQueryService.cs
@@ -7,7 +7,7 @@ namespace Fleans.Application;
 public interface IWorkflowQueryService
 {
     Task<InstanceStateSnapshot?> GetStateSnapshot(Guid workflowInstanceId);
-    Task<IReadOnlyList<ProcessDefinitionSummary>> GetAllProcessDefinitions();
+    Task<int> GetMaxVersionByKeyAsync(string processDefinitionKey);
     Task<PagedResult<ProcessDefinitionSummary>> GetAllProcessDefinitions(PageRequest page);
     Task<PagedResult<ProcessDefinitionGroup>> GetProcessDefinitionGroups(PageRequest page);
     Task<PagedResult<WorkflowInstanceInfo>> GetInstancesByKey(string processDefinitionKey, PageRequest page);
--- a/src/Fleans/Fleans.Persistence.Tests/WorkflowQueryServiceTests.cs
+++ b/src/Fleans/Fleans.Persistence.Tests/WorkflowQueryServiceTests.cs
@@ -254,7 +254,7 @@ public class WorkflowQueryServiceTests : PersistenceTestBase
     [DataTestMethod]
     [DataRow(PersistenceProvider.Sqlite)]
     [DataRow(PersistenceProvider.Postgres)]
-    public async Task GetAllProcessDefinitions_ReturnsAll_OrderedByKeyThenVersion(PersistenceProvider provider)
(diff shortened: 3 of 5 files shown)


nightBaker (commented):
## Code Review — #694 fix(#611) targeted max-version query

Clean implementation of Option C from the approved Design & Plan v3. Tests pass (154 SQLite persistence, 4 ProcessDefinitionGrain, 0 failures), diff is minimal (5 files, 95+/31−), no CLAUDE.md violations.

### Verified

- **Behavioral equivalence** between old and new Editor logic:
  - never-deployed key: old `existing==null → nextVersion=1 → "v1"`; new `max=0 → 0+1=1 → "v1"` ✓
  - 3 deployments (v1,v2,v3): old `existing.Version=3 → nextVersion=4 → "v4"`; new `max=3 → 3+1=4 → "v4"` ✓
- **`Select(d => (int?)d.Version).MaxAsync() ?? 0`** correctly handles empty result sets (EF Core returns null, `?? 0` yields 0). Confirmed by `GetMaxVersionByKeyAsync_ReturnsZero_WhenEmpty` and `…_WhenUnknownKey`.
- **Paged-overload test rewrite** correctly uses `Sorts: "processDefinitionKey,version"` to preserve the ordering coverage the removed no-arg method used to guarantee. Sieve attribute mapping for both fields verified at `ApplicationSieveProcessor.cs:28-30` (`.CanSort()`).
- **API-break scope**: `grep -rn "GetAllProcessDefinitions()" src/` returns zero hits after the change. `IWorkflowQueryService` is not in the published NuGet stack (lives in `Fleans.Application`, not `Fleans.Application.Abstractions`) — no external consumer impact.

### 🟢 Minor (non-blocking, can be addressed in follow-up)

1. **No `CancellationToken` parameter on `GetMaxVersionByKeyAsync`.** Most methods on `IWorkflowQueryService` also lack a CT parameter, so this matches existing convention — but `GetRegisteredEventsAsync(CancellationToken ct = default)` is the precedent for new query methods. Adding `CancellationToken ct = default` would be future-proof for caller cancellation (especially on the Editor's Blazor circuit if it disconnects mid-call). Defer to a separate sweep if maintainer wants consistency across all query methods at once.

2. **`BpmnXml` projection-pushdown claim is supported by the design, not explicitly verified.** The plan asserted that `Select(d => (int?)d.Version)` short-circuits `BpmnXml` materialization. EF Core docs confirm projection pushdown, but this PR doesn't capture the generated SQL anywhere. A trivial verification step (e.g. `dotnet ef dbcontext optimize` or a debug log inspection) would close the loop — not worth gating the PR on.

Issue #628, opened 2026-05-19
ExpandoObjectSurrogateConverter not available to external plugin hosts — CodecNotFoundException on IWorkflowInstanceCallback


nightBaker:
## Problem

`Fleans.Domain.Surrogates.ExpandoObjectSurrogateConverter` (the surrogate that makes `System.Dynamic.ExpandoObject` Orleans-serializable) lives in `Fleans.Domain` — the engine-internal package that external plugin hosts deliberately don't reference. The leaf-package design at v0.2.0 (per CLAUDE.md *Plugin-author NuGet stack*) means a plugin host's closure is `Fleans.Worker` → `Fleans.Application.Abstractions` → `Fleans.Domain.Abstractions`, with `Fleans.Domain` excluded by design.

But `IWorkflowInstanceCallback` (the narrow interface the plugin host calls back into the engine through) takes `ExpandoObject variables` as a parameter. When the plugin host's silo tries to materialise a proxy or copy the argument, Orleans serialization can't find a copier for the type.

## Repro

External plugin host completes a custom task; on the callback to `IWorkflowInstanceCallback.CompleteActivity`:

```
warn: Orleans.Streams.StreamProvider[361474169]
      Failed to deliver message to consumer on <id> for stream
      StreamProvider/events.ExecuteCustomTaskEvent.hello-world/<workflowInstanceId>, may retry.

Orleans.Serialization.CodecNotFoundException: Could not find a copier for type System.Dynamic.ExpandoObject.
   at Orleans.Serialization.Serializers.CodecProvider.ThrowCopierNotFound(Type type)
   at Orleans.Serialization.Serializers.CodecProvider.GetDeepCopier[T]()
   at Orleans.Serialization.ServiceCollectionExtensions.CopierHolder`1.get_Value()
   at Proxy_IWorkflowInstanceCallback(Object, GrainReferenceShared, IdSpan)
   at Orleans.GrainReferences.GrainReferenceActivatorProvider.GrainReferenceActivator.CreateReference(GrainId grainId)
   at Orleans.GrainReferences.GrainReferenceActivator.CreateReference(GrainId grainId, GrainInterfaceType interfaceType)
   at Orleans.GrainFactory.GetGrain(Type interfaceType, IdSpan grainKey, String grainClassNamePrefix)
```

[...]

Issue #627, opened 2026-05-19
AddFleansPluginHost missing CorePlacementStrategy registration — external plugin hosts can't register with catalog


nightBaker:
## Problem

`Fleans.Worker.Hosting.PluginHostExtensions.AddFleansPluginHost` only registers `WorkerPlacementStrategy`:

```csharp
public static ISiloBuilder AddFleansPluginHost(this ISiloBuilder siloBuilder, IConfiguration configuration)
{
    var role = ValidatePluginRole(configuration["Fleans:Role"]);
    var siloName = BuildSiloName(role, Environment.MachineName, Guid.NewGuid());

    siloBuilder.Configure<Orleans.Configuration.SiloOptions>(o => o.SiloName = siloName);
    siloBuilder.AddPlacementDirector<WorkerPlacementStrategy, WorkerPlacementDirector>();
    return siloBuilder;
}
```

This contradicts the established invariant from CLAUDE.md *Core / Worker role split*:

> **Role-aware placement directors. Every silo registers both so a Core-only silo can still route worker grains to its Worker siblings (and vice versa).**

External plugin hosts built from `fleans-custom-worker-example` need to **target** Core-placed grains (specifically `CustomTaskCatalogGrain`, which is `[CorePlacement]`) to register their plugins with the catalog UI. Without `CorePlacementStrategy` registered locally, Orleans throws when the plugin host tries to call into `customtaskcatalog`.

## Repro

External plugin host on the same Orleans cluster as a Fleans engine deployment. On plugin registration:

```
warn: Fleans.Worker.CustomTasks.CustomTaskPluginRegistrar[9331]
      Catalog Register failed for 'hello-world' (attempt 1/4); retrying in 00:00:02
      System.Collections.Generic.KeyNotFoundException: Could not resolve placement strategy
        CorePlacementStrategy for grain type customtaskcatalog.
         at Orleans.Runtime.Placement.PlacementStrategyResolver.TryGetNonDefaultPlacementStrategy(GrainType grainType, PlacementStrategy& strategy)
         at Orleans.Runtime.Placement.PlacementStrategyResolver.GetPlacementStrategyInternal(GrainType grainType)
         ...
         at Orleans.Runtime.Placement.PlacementService.PlacementWorker.GetOrPlaceActivationAsync(Message firstMessage)
```

After 4 retries:

```
warn: Fleans.Worker.CustomTasks.CustomTaskPluginRegistrar[9332]
[...]

PR #704, opened 2026-05-29
fix(#703): disable EndpointOptions.GatewayPort on plugin host silos

nightBaker:
Closes #703.

## Summary

`AddFleansPluginHost` was leaving Orleans's gateway port at the default, so plugin host silos advertised themselves as **client gateways** when joining the cluster via Redis clustering. `Fleans.Web` is an Orleans client (`UseOrleansClient`) and round-robins among advertised gateways. When the client landed on a plugin gateway and forwarded an engine grain call through it, the gateway threw the exact error reported in #703:

> `Unable to load Fleans.Application.Grains.IWorkflowInstanceGrain, Fleans.Application from assembly Fleans.Application`

— because the plugin silo's assembly load context intentionally lacks `Fleans.Application` (the whole point of the leaf-package design).

The one-line fix sets `EndpointOptions.GatewayPort = 0` inside `AddFleansPluginHost`, telling Orleans the silo does NOT advertise as a client gateway. Silo-to-silo traffic (the `SiloPort`) is unaffected — the silo still joins the cluster, participates in placement, and serves custom-task stream subscriptions for the plugin grain types it hosts.

## End-to-end verified

I reproduced the bug locally against the v0.5.0 release bundle + `fleans-custom-worker-example`, then applied this fix as a runtime override (`Orleans__Endpoints__GatewayPort=0`) on the running stack:

- Without the override: intermittent "Unable to load type" errors when starting workflows via the management Web.
- With the override applied: 10/10 starts succeed; HelloWorld plugin still activates on the plugin silo as designed; workflow completion + return-variables round-trip cleanly.

Workaround stop-gap for users on pre-fix Fleans versions documented in a parallel PR against [`fleans-custom-worker-example`](https://github.com/nightBaker/fleans-custom-worker-example).

## Test plan

- [x] `dotnet test --filter "PluginHostPlacementTests"` — 6/6 pass (1 new regression guard + 5 existing).
- [x] Build clean (0 errors, 132 pre-existing warnings).
- [ ] Manual: cycle the running v0.5.0 + custom-worker stack with the engine-fix applied (remove the env override, rebuild custom-worker image with this branch's `Fleans.Worker` package). I left the env-var workaround on the local stack for now — happy to do this cycle once the PR is approved.

## Related

- #627 (`AddFleansPluginHost` missing `CorePlacementStrategy` — landed pre-v0.5.0). Same surface area as this fix; both were latent multi-silo-cluster bugs that only surface when a plugin host is in the cluster.
- #628 (`ExpandoObjectSurrogateConverter` moved to `Fleans.Domain.Abstractions` — landed pre-v0.5.0).

### DIFF
--- a/src/Fleans/Fleans.Application.Tests/Placement/PluginHostPlacementTests.cs
+++ b/src/Fleans/Fleans.Application.Tests/Placement/PluginHostPlacementTests.cs
@@ -81,6 +81,30 @@ public class PluginHostPlacementTests
     }
 
     [TestMethod]
+    public void AddFleansPluginHost_DisablesGatewayPort()
+    {
+        // #703 regression guard. Plugin host silos must NOT advertise as Orleans client
+        // gateways: their assembly load context intentionally lacks Fleans.Application /
+        // Fleans.Domain / Fleans.Persistence. An engine grain call (e.g. IWorkflowInstanceGrain)
+        // forwarded *through* a plugin gateway throws "Unable to load type" at the gateway.
+        // Removing the EndpointOptions.GatewayPort = 0 line from AddFleansPluginHost MUST
+        // make this test fail.
+        var configuration = new ConfigurationBuilder()
+            .AddInMemoryCollection(new Dictionary<string, string?> { ["Fleans:Role"] = "Plugin" })
+            .Build();
+
+        var builder = new FakeSiloBuilder(configuration);
+        builder.AddFleansPluginHost(configuration);
+
+        using var provider = builder.Services.BuildServiceProvider();
+        var endpoints = provider.GetRequiredService<Microsoft.Extensions.Options.IOptions<Orleans.Configuration.EndpointOptions>>().Value;
+
+        Assert.AreEqual(0, endpoints.GatewayPort,
+            "AddFleansPluginHost must set EndpointOptions.GatewayPort = 0 — plugin silos lack " +
+            "Fleans.Application and cannot serve as Orleans client gateways for engine grains (#703).");
+    }
+
+    [TestMethod]
     public void AddFleansPluginHost_RegistersBothPlacementDirectors()
     {
         // #627 regression guard. Removing either AddPlacementDirector<> line from
--- a/src/Fleans/Fleans.Worker/Hosting/PluginHostExtensions.cs
+++ b/src/Fleans/Fleans.Worker/Hosting/PluginHostExtensions.cs
@@ -16,6 +16,11 @@ public static class PluginHostExtensions
         var siloName = BuildSiloName(role, Environment.MachineName, Guid.NewGuid());
 
         siloBuilder.Configure<Orleans.Configuration.SiloOptions>(o => o.SiloName = siloName);
+        // Plugin hosts must not advertise as Orleans client gateways: their assembly load context
+        // intentionally lacks Fleans.Application / Fleans.Domain / Fleans.Persistence, so any
+        // engine grain call (e.g. IWorkflowInstanceGrain) forwarded *through* a plugin gateway
+        // throws "Unable to load type" at the gateway. See #703.
+        siloBuilder.Configure<Orleans.Configuration.EndpointOptions>(o => o.GatewayPort = 0);
         siloBuilder.AddPlacementDirector<CorePlacementStrategy, CorePlacementDirector>();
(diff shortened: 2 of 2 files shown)


nightBaker (commented):
## Code Review — fix(#703): disable EndpointOptions.GatewayPort on plugin host silos

Self-review (I authored this PR). Ran with extra rigor to compensate for the bias.

### Scope verification (CLAUDE.md compliance)

- ✅ Production change is minimal (+5 lines, 1 `Configure` call) and lives in the right file (`PluginHostExtensions.cs`) — same place `AddFleansPluginHost` already sets `SiloOptions.SiloName`.
- ✅ Regression-guard test added next to the existing `#627` test (`AddFleansPluginHost_RegistersBothPlacementDirectors`), matching the file's established pattern.
- ✅ Comment in `PluginHostExtensions.cs` explains the "why" with a `#703` cross-reference. The comment is exactly the kind that survives — references a specific failure mode, not "what the code does."
- ✅ No new public API, no breaking changes for users not setting `GatewayPort` explicitly.

### Critical checks

1. **Combined role.** `ValidatePluginRole` accepts both `Plugin` and `Combined` for plugin hosts. The fix disables the gateway for both. Is this correct for `Combined`?
   - In the **engine** (Api/WorkerHost), `Combined` means "single silo hosts Core + Worker grains" — those silos *do* have `Fleans.Application` and *should* be gateways. But those silos don't use `AddFleansPluginHost` (they have their own Program.cs setup, verified by `grep`).
   - In `AddFleansPluginHost`, `Combined` means a plugin host that's also acting in a Worker-like capacity for placement — but it still doesn't load `Fleans.Application` (leaf-package design). Disabling the gateway is still correct.
   - Edge case: a user who *adds* `Fleans.Application` to their plugin host's project refs (voluntarily abandoning the leaf-package design). For that hypothetical user, the gateway is now disabled even though they could serve gateway traffic. Mitigated by Orleans's options pattern: they can override with `siloBuilder.Configure<EndpointOptions>(o => o.GatewayPort = X)` *after* `AddFleansPluginHost` (last-wins). Worth mentioning in docs but not blocking.

2. **Orleans `GatewayPort = 0` semantics.** Empirically verified: in the local v0.5.0 + custom-worker reproduction, setting `Orleans__Endpoints__GatewayPort=0` stopped the intermittent "Unable to load type" errors immediately. PR's claim that silo-to-silo traffic (SiloPort) is unaffected is consistent with Orleans's `EndpointOptions` design: `SiloPort` and `GatewayPort` are independent endpoints.

3. **Backward compatibility.** Existing plugin host deployments without an explicit `GatewayPort` setting will now have it set to 0. The only thing that was happening with the default port was the bug we're fixing. Strict improvement for the documented use case.

4. **Test correctness.** The test uses `FakeSiloBuilder` (already established in the file for `#627`), runs `AddFleansPluginHost`, builds a `ServiceProvider`, and inspects `IOptions<EndpointOptions>.Value`. This triggers all `Configure<EndpointOptions>` callbacks registered by the SUT. The fully-qualified `Microsoft.Extensions.Options.IOptions` + `Orleans.Configuration.EndpointOptions` is verbose but matches the file's existing `Orleans.Configuration.SiloOptions` style (no aliasing).

5. **Test isolation.** `using var provider` ensures the `ServiceProvider` is disposed after the test — no leak across tests.

6. **CI.** build ✓ · e2e ✓ · pg-tests ✓. Local: 6/6 PluginHostPlacementTests pass.

7. **PR description.** Identifies the root cause precisely (Orleans client gateway round-robin), reproduces the failure deterministically (v0.5.0 release bundle + custom-worker-example), explains the fix mechanism, and notes the workaround for current users. Test plan checks the right boxes.

### 🟢 Minor

**Operator override path could be documented.** Operators with the rare advanced use case (plugin host that *does* have engine assemblies loaded, e.g., a Combined host with explicit `Fleans.Application` reference) can override by calling `siloBuilder.Configure<EndpointOptions>(o => o.GatewayPort = X)` AFTER `AddFleansPluginHost` — Orleans's options pattern applies these in registration order. Worth a sentence in the XML doc on the extension method, or a follow-up to the `docs/conventions/placement-and-roles.md` page. Not blocking.

### Verdict

No blocking issues. CI green. Regression test pins the fix. OK to promote.

PR #707, opened 2026-06-02
security: fail-closed when running Production without authentication

spqr911:
## Summary

Refuse to start `Fleans.Api` and `Fleans.Web` in Production if the `Authentication` section is empty. Today both processes silently fall into anonymous mode — every controller and every Blazor page becomes public, and the Orleans Dashboard joins them. This PR makes that misconfiguration loud at process start rather than a quiet hole at runtime.

No dev-loop change: `dotnet run --project Fleans.Aspire` (which sets `ASPNETCORE_ENVIRONMENT=Development`) keeps working exactly as before. Operators who deliberately want an unauthenticated install in Production must set `ASPNETCORE_ENVIRONMENT=Staging` (or supply credentials).

## Problem

Both `Fleans.Api/Program.cs` and `Fleans.Web/Program.cs` wrap their entire auth pipeline in `if (authEnabled)`:

- `AddJwtBearer` / `AddOpenIdConnect` registration
- `SetFallbackPolicy(RequireAuthenticatedUser())`
- `UseAuthentication` / `UseAuthorization` middleware

With `Authentication:Authority` empty (the default in `appsettings.json` and in the Helm chart's `values.yaml`), none of these run. Controllers under `Fleans.Api/Controllers/` carry **no `[Authorize]` attributes** (verified via grep), so anonymous callers can reach:

- `POST /Definitions/deploy` — load arbitrary BPMN, including `<scriptTask scriptFormat="csharp">` which `DynamicExpresso` will execute in-process on the worker silo
- `POST /Execution/start`, `POST /Execution/message`, `POST /Execution/signal`
- `POST /UserTasks/{id}/claim` — combined with `BodyUserGroupResolver`, the body's `userGroups` is taken on trust
- `POST /Definitions/disable`, `enable`

The Web project is worse: per the design rule in `CLAUDE.md`, Blazor pages call Orleans grains **directly** instead of going through the API, so "fix it at the API perimeter" doesn't help. And `app.MapOrleansDashboard("/dashboard")` (line 200 of `Fleans.Web/Program.cs`) sits **outside** the `if (authEnabled)` block — the explicit gate above it (lines 186-196) is itself conditional on `authEnabled`, so when auth is off the gate isn't installed.

Cumulative effect: a stock `helm install fleans` with no `auth.authority` value lands a public BPMN-deploy endpoint plus a public Orleans Dashboard plus a public Blazor admin UI on whatever network the cluster is on.

## Industry comparison

**Camunda 7 / Zeebe** ship the same fail-open posture by default (REST API without authentication when no Identity provider is configured). This has produced a documented history of "Camunda exposed on the internet" advisories. Camunda 8 SaaS hardened by requiring API tokens for every request; self-hosted Camunda still requires the operator to opt in to auth. Fleans inherits this BPMN-engine family trait.

**Temporal** has a similar "self-hosted dev mode is open" trait (`temporal server start-dev` listens without auth), but with two differences: (a) its server never executes user-supplied workflow code, so the worst case is data exfiltration rather than RCE; (b) the production Temporal Helm charts and Temporal Cloud both default to mTLS-required. Fleans here moves one step toward the Temporal posture: dev mode stays easy, Production fails closed.

## Change

Two new pre-`builder.Build()` checks, one in each `Program.cs`:

\`\`\`csharp
if (!authEnabled && builder.Environment.IsProduction())
{
    throw new InvalidOperationException(
        \"Fleans.<X> refuses to start in Production without authentication. \" +
        \"Set 'Authentication:Authority' …\");
}
\`\`\`

No new packages, no new abstractions, no behaviour change in Development/Staging.

## What this PR deliberately does **not** do

- **Add `[Authorize]` to every controller.** That's a follow-up PR and needs a regression pass on the E2E suite.
- **Touch `Fleans.Mcp`.** MCP has no auth wiring at all today; the right fix is a follow-up PR that adds an auth layer or restricts the listener to loopback. Adding only a Production guard there would leave the dev-mode surface wide open and provide false confidence.
- **Touch `Fleans.WorkerHost`.** It only maps `/health` and `/alive`; no business-facing HTTP routes.
- **Force-enable auth in Development.** A common dev loop is `dotnet run --project Fleans.Aspire` against local SQLite + Memory streams; requiring an IdP there would be hostile.

## Test plan

Locally verified (.NET 10.0.108 SDK, branch built clean, 447 unit tests pass):

- [x] `ASPNETCORE_ENVIRONMENT=Production`, no Authentication config → both processes throw \`InvalidOperationException\` at startup with a clear message pointing at the missing config key. Throw fires before Orleans / DB / Redis are touched.
- [x] `ASPNETCORE_ENVIRONMENT=Production`, `Authentication__Authority` set → check passes, process proceeds to normal startup (then fails on missing Redis in my local sandbox, which is expected).
- [x] `ASPNETCORE_ENVIRONMENT=Development` (default in launchSettings.json) → no behaviour change.

What I could not validate locally (no infra):

- [ ] CI build + existing E2E pass — relying on the PR pipeline.
- [ ] Helm install with empty `auth.authority` and `ASPNETCORE_ENVIRONMENT=Production` → pod CrashLoopBackOff with the new message in `kubectl logs`.

## Follow-ups (separate PRs)

- Move `MapOrleansDashboard` inside the `if (authEnabled)` block in \`Fleans.Web/Program.cs\` so even Development mode stops exposing the cluster dashboard anonymously.
- Wire authentication into `Fleans.Mcp`.
- Add `[Authorize]` baseline to controllers so a future regression in \`Program.cs\` can't reopen the hole.

### DIFF
--- a/src/Fleans/Fleans.Api/Program.cs
+++ b/src/Fleans/Fleans.Api/Program.cs
@@ -30,6 +30,22 @@ builder.AddServiceDefaults();
 // Authentication — opt-in: only enabled when Authentication:Authority is configured
 var authAuthority = builder.Configuration["Authentication:Authority"];
 var authEnabled = !string.IsNullOrEmpty(authAuthority);
+
+// Fail-closed in Production: an empty Authentication:Authority in non-Development
+// environments would leave every controller anonymous (none of them carry [Authorize]
+// individually — the fallback policy is the only protection, and that is only
+// registered inside `if (authEnabled)` below). Refuse to start instead of silently
+// shipping an open admin surface. Operators who knowingly want an unauthenticated
+// deployment must opt in by setting ASPNETCORE_ENVIRONMENT=Development or Staging.
+if (!authEnabled && builder.Environment.IsProduction())
+{
+    throw new InvalidOperationException(
+        "Fleans.Api refuses to start in Production without authentication. " +
+        "Set 'Authentication:Authority' (OIDC issuer URL) and 'Authentication:Audience'. " +
+        "To run unauthenticated for local/dev use, set ASPNETCORE_ENVIRONMENT to " +
+        "Development or Staging.");
+}
+
 if (authEnabled)
 {
     builder.Services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
--- a/src/Fleans/Fleans.Web/Program.cs
+++ b/src/Fleans/Fleans.Web/Program.cs
@@ -56,6 +56,21 @@ var authAuthority = builder.Configuration["Authentication:Authority"];
 var authClientId = builder.Configuration["Authentication:ClientId"];
 var authEnabled = !string.IsNullOrEmpty(authAuthority) && !string.IsNullOrEmpty(authClientId);
 
+// Fail-closed in Production: with auth off, the fallback policy is never registered
+// (see `if (authEnabled)` block below) AND the Orleans Dashboard mapping at the bottom
+// of this file is unconditional, so an empty Authentication section in Production
+// would publish an open admin UI + cluster dashboard. Refuse to start instead.
+// Operators who knowingly want an unauthenticated deployment must opt in by setting
+// ASPNETCORE_ENVIRONMENT=Development or Staging.
+if (!authEnabled && builder.Environment.IsProduction())
+{
+    throw new InvalidOperationException(
+        "Fleans.Web refuses to start in Production without authentication. " +
+        "Set 'Authentication:Authority' (OIDC issuer URL) and 'Authentication:ClientId'. " +
+        "To run unauthenticated for local/dev use, set ASPNETCORE_ENVIRONMENT to " +
+        "Development or Staging.");
+}
+
 builder.Services.AddSingleton(new AuthOptions(authEnabled, authAuthority ?? "", authClientId ?? ""));
 
 if (authEnabled)


nightBaker (changes requested):
Thanks for the thorough writeup — the threat analysis here is genuinely good, and the inventory of the anonymous surface (public `/Definitions/deploy` with `csharp` script execution, the unconditional `MapOrleansDashboard`, the `BodyUserGroupResolver` trust) is accurate. I'm going to **decline this change**, though, and I want to explain the reasoning so it doesn't read as a dismissal of the security concern.

## Motivation: getting-started simplicity is a deliberate product decision

Fleans is intentionally **zero-auth by default**. The first-run experience is the priority: a new user should be able to run

```
docker compose up -d      # released bundle
helm install fleans ...   # stock chart
```

and land on a working admin UI in minutes, with no IdP to stand up first. That's a conscious trade-off — unauthenticated-by-default — and it's documented as such (`authentication.md`: *"disabled by default … runs fully unauthenticated"*). I want to keep it.

## Why this PR conflicts with that

The guard keys on `builder.Environment.IsProduction()`, but **none of the shipped artifacts set `ASPNETCORE_ENVIRONMENT`** — not the Helm chart templates, not the Aspire compose/k8s publish, not the released compose bundle. .NET defaults to `Production` when the variable is unset, and all of them ship the `Authentication` section empty by default. So the net effect isn't "fail closed on a misconfigured prod install" — it's:

- `docker compose up -d` from the Quick Start bundle → every service throws at startup. The documented 5-minute onboarding is broken end-to-end.
- stock `helm install fleans` → CrashLoopBackOff.

The PR's threat model assumes `Production` is an explicit operator choice; in practice it's the default for every release artifact, so the guard lands on the happy path rather than on the people who deliberately exposed an open install. That's the opposite of the simplicity I want to protect.

## The deeper reason I don't want the hard gate

Secure-by-default for this engine would mean shipping credentials or forcing an IdP into the quick-start — which is exactly the friction we're trading away on purpose. I'd rather keep the open default and make the *production-hardening path* obvious and easy, than make the default path refuse to boot.

## What I'd welcome instead

A **docs-only** PR: add a "Securing a production deployment" section to `website/src/content/docs/reference/authentication.md` (and a cross-link from the self-host Docker Compose / Helm guides) that walks operators through enabling OIDC — the opt-in wiring already exists. That captures all the value of this analysis without breaking the default experience.

The follow-ups you listed (moving `MapOrleansDashboard` inside the auth block, wiring auth into `Fleans.Mcp`, a `[Authorize]` baseline) are also reasonable as independent issues — they harden the *opt-in* path without changing the default.

Closing on those grounds. Really appreciate the depth here.

PR #726, opened 2026-06-11
fix(#657): durable pending-external-events queue + op-id dedup ledger

nightBaker:
## Summary

Closes #657. Makes the `WorkflowInstance` pending-external-events queue durable so it survives forced grain deactivation, and adds an op-id dedup ledger so caller-retried sibling-grain RPCs are never double-applied.

The bug: `WorkflowInstance` held child-completed/failed/escalation/signal events in an **in-memory** `ConcurrentQueue` (`_pendingExternalEvents`) drained on a grain timer. On forced deactivation (memory pressure / eviction / shutdown) the queue **and** the awaiting `TaskCompletionSource`s were silently lost.

Implements the approved **v5** design (issue comments) + the round-5 review guidance. Five rounds of design review converged on this; round-4 fixed the escalation op-id threading, round-5 confirmed and added the domain-purity guidance applied here.

## What changed

- **Durable queue** — each enqueue persists a `PendingEventEnqueued` domain event via the JournaledGrain log; `OnActivateAsync` rebuilds the in-memory queue (with fresh TCSes) from `State.PendingOperations`. The drain path emits `PendingEventDrained` (+ `PendingEventApplied` for retried ops).
- **Op-id dedup ledger** (`State.PendingOperations` / `State.AppliedOperations`) on the **three caller-retried** methods (`OnChildWorkflowCompleted`/`OnChildWorkflowFailed`/`OnChildEscalationRaised`): applied → return persisted outcome; in-flight → await existing TCS; fresh → durably enqueue. This is documented as **guard #7** in CLAUDE.md *Stream-redelivery idempotency*. `AppliedOperations` is GC'd on terminal state (guard (2)'s `IsCompleted` covers post-completion retries).
- **Signal path is never retried** (`SignalCorrelationGrain` consumes the subscription before fire-and-forget delivery), so it persists with a fresh-Guid key and **no** dedup. `HandleBoundarySignalFired` is test-only and left untouched.
- **Escalation discriminator** — the originating throw's `activityInstanceId` is carried as a raw `Guid EscalationInstanceId` on `NotifyParentEscalationRaisedEffect`, threaded through `OnChildEscalationRaised → PendingChildEscalationRaised → HandleChildEscalationRaised`, and **propagated unchanged** at the re-escalation emit so a retried hop dedups to the same op-id. Per round-5 guidance, the `child-escalation:{…}` **string** is formatted only in the application layer (`GrainFactoryRetryExtensions`); the domain carries only the Guid.
- **Caller side** — the three effect-handler call sites are wrapped in a new `IGrainFactory.CallWithRetry` extension (exp-backoff on transient `OrleansException`).

## Key decisions / points for review

1. **Enqueue persistence path.** The `[AlwaysInterleave]` enqueue methods persist via a serialized (`_enqueueLock`) `RaiseEvent(PendingEventEnqueued)` + `ConfirmEvents` under the existing `_draining` guard (apply-then-persist-skip, the same shape as `DrainAndRaiseEvents`), rather than routing through the aggregate effect-drain cycle (which the interleaved callers can't enter). Replay re-applies via `WorkflowExecution.Apply`. Worth a careful look that this composes with the six existing redelivery guards.
2. **Durability scope.** `PendingOperations`/`AppliedOperations` are rebuilt from the event log on activation (same pattern as `TransactionOutcomes`) and excluded from the EF projection. They survive the ungraceful-deactivation path #657 is about (events replay after the last snapshot); the graceful path already drains the queue. Tests assert journal persistence + reactivation idempotency.
3. **`PendingChildFailedPayload` carries `ExceptionMessage` (string)**, not the full `Exception` (not serializable across the journal). On reactivation-replay a re-queued child failure reconstructs from the message only — exception type/stack fidelity is lost. Flagging in case richer failure payload is wanted.

## Tests

New `PendingEventsDurabilityTests.cs` (8): journal-persist+reactivation, duplicate-op-id short-circuit, escalation retry returns persisted result, reactivation replays with fresh TCS, terminal-state GC, repeated non-interrupting signal applies each fire, repeated same-code distinct-throw escalation applies each, and A→B→C re-escalation propagates the same op-id (dedup across hops).

- `dotnet build`: green.
- New tests: 8/8 pass (controller-verified).
- Targeted regression (Escalation / Signal / ChildWorkflow / SignalBoundary / CallActivity): green (116 tests).

## How to test

`cd src/Fleans && dotnet test --filter "FullyQualifiedName~PendingEventsDurability"` and the targeted regression filter above.


nightBaker (commented):
-

nightBaker (commented):
## Code Review — fix(#657): durable pending-external-events queue + op-id dedup ledger

(Re-submitting: the prior approve review's body was lost to a `gh pr review --body -` quirk — `--body -` sets the literal string `-` rather than reading stdin; the marker never landed.)

Reviewed all changed files against CLAUDE.md and the v5 design. The implementation is correct, replay-safe, and consistent with the codebase's event-sourcing patterns. No blocking issues.

**Verified:**
- **Durability is journal-backed.** Enqueue persists `PendingEventEnqueued` via `RaiseEvent` under the `_draining` guard (apply-then-persist-skip, same shape as `DrainAndRaiseEvents`); test 1 asserts the `PendingEvent*` events are in `IEventStore` and survive `ForceAllGrainDeactivation()`. No `WriteStateAsync` in `WorkflowInstance`.
- **Terminal GC is replay-safe.** `AppliedOperations.Clear()` lives in `WorkflowInstanceState.Complete()`/`Cancel()`, invoked from the `WorkflowCompleted`/`Cancelled` Apply handlers — re-clears deterministically on replay, after the `PendingEventApplied` events that populated it.
- **Op-id dedup (guard #7)** is a clean 3-branch state machine on the three caller-retried methods; signal path persists with a fresh-Guid key and no dedup; documented in CLAUDE.md.
- **Escalation discriminator** threaded end-to-end as a raw `Guid` (origin `:1002` → effect → `OnChildEscalationRaised` → `PendingChildEscalationRaised` → `HandleChildEscalationRaised` → propagated `:2045`); op-id string formatted only in the application layer — domain-purity respected.
- **`CallWithRetry`** bounded (final attempt rethrows); deterministic op-id helpers; child-id from `context.WorkflowInstanceId`.
- Build green; new `PendingEventsDurabilityTests` 8/8 pass; targeted regression green.

### 🟡 Important
1. **`PendingChildFailedPayload` carries only `ExceptionMessage`** (not the full `Exception` — not journal-serializable). On reactivation-replay of a re-queued child failure the exception is reconstructed as `new Exception(message)`, losing type/stack. Confirm no downstream failure handling branches on exception type.

### 🟢 Minor
1. Durability tests are plumbing-level (`"absent-call"`/`"absent-host"` ids), exercising the pending-events layer in isolation rather than a real child-workflow→parent completion across deactivation. An end-to-end test would strengthen confidence (the manual-regression Testing stage partially covers this).
2. `CallWithRetry` catches only `OrleansException` — fine because the dedup ledger makes retry safe, but worth a comment.
3. The direct `RaiseEvent`+`ConfirmEvents` enqueue path (outside the aggregate drain cycle, for the `[AlwaysInterleave]` callers) is well-commented; flagged only so it reads as deliberate.

No blocking issues. OK to promote (pending CI green).

PR #736, opened 2026-06-16
feat(streaming-kafka): mTLS / client-cert config (#681)

nightBaker:
## Summary

- Adds four SSL properties to `KafkaStreamingOptions`: `SslCaLocation`, `SslCertificateLocation`, `SslKeyLocation`, `SslKeyPassword`
- Restructures `ApplySecurity` to fix a dead-code bug: the early-return that blocked `SecurityProtocol=Ssl` from reaching the SSL block is replaced with a conditional SASL block, so control always falls through to the new SSL section
- Three fail-fast validation throws: SSL paths with incompatible protocol, unpaired cert/key, password without key
- OS-trust-store WARNING (EventId 11100) at each adapter call site when no `SslCaLocation` is set with an SSL protocol
- 11 new SSL unit tests in `KafkaClientConfigExtensionsSslTests`
- `docs/conventions/streaming.md` updated with mTLS subsection; `observability-eventids.md` updated with range 11100–11199
- Manual test plan at `tests/manual/67-kafka-mtls/test-plan.md`

Closes #681

## Test plan

- [ ] `dotnet test --filter "Kafka"` — 34 tests pass (13 existing SASL + 11 new SSL)
- [ ] Manual test plan: `tests/manual/67-kafka-mtls/test-plan.md` — OS-trust-store WARN, CA-only mode, full mTLS, misconfiguration rejection

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #737, opened 2026-06-18
feat(streaming-kafka): AWS MSK IAM authentication via OAuthBearer (#682)

### LINKED ISSUE #682: [feat:streaming-kafka] AWS MSK IAM authentication (OAUTHBEARER provider hook)
Tracking: #474 (Tier 🔴 — won't connect to managed brokers)

## What

Wire AWS MSK IAM token provider as an OAUTHBEARER mechanism. Implemented as a separate optional package `Fleans.Streaming.Kafka.AwsMsk` (in-repo, per maintainer decision).

## Acceptance Criteria

- [ ] New project `Fleans.Streaming.Kafka.AwsMsk` created in-repo
- [ ] AWS MSK IAM token provider wired as OAUTHBEARER mechanism via `OAuthBearerTokenProvider` from sub-issue A
- [ ] Workload-identity supported via `AWSSDK.Core` default credential chain
- [ ] `AWSSDK.Core` is a dependency of the AwsMsk package only (not the core Kafka adapter)
- [ ] Extension method `AddKafkaStreamingWithMskIam()` or similar for easy registration
- [ ] Unit tests for token provider wiring
- [ ] Documentation: usage example in `docs/conventions/streaming.md`

## Dependencies

- Sub-issue A (#680) must land first (provides `OAuthBearerTokenProvider` hook)

## Related

- #474 (parent tracking issue)

nightBaker:
## Summary

- New optional in-repo package `Fleans.Streaming.Kafka.AwsMsk` that wires AWS SigV4 token generation into Kafka's OAUTHBEARER SASL mechanism via a one-line silo-builder extension
- IAM credentials come from the standard AWS workload identity chain (EC2 instance profile, EKS Pod Identity / IRSA, ECS task role) — no static secrets in configuration
- Region resolves: explicit param → `AWS_REGION` env var → `AWS_DEFAULT_REGION` env var → throws `InvalidOperationException` at startup before any broker connection

## Key decisions

- **New package instead of modifying core**: `Fleans.Streaming.Kafka.AwsMsk` is a separate in-repo assembly, keeping `AWS.MSK.Auth` out of the base Kafka streaming package. Operators that don't use MSK pay zero dependency cost.
- **`async void` callback is deliberate**: Confluent.Kafka's `SetOAuthBearerTokenRefreshHandler` takes `Action<IClient, string>`, so async token fetch must be fire-and-forget. Failure signals via `OAuthBearerSetTokenFailure`; a nested try/catch swallows any throw from that call when the client is disposed.
- **`AddOptions<T>.Configure<ILoggerFactory>()`**: Used to resolve `ILogger` at options-resolution time (not at `AddKafkaStreamingWithMskIam` call time), when DI is fully built.
- **EventIds 11200–11299** allocated per `docs/conventions/observability-eventids.md` (11100–11199 was already taken by PR #736).

## Testing

4 unit tests in `Fleans.Streaming.Kafka.AwsMsk.Tests` cover:
- Security options are wired correctly when OAuthBearer callback is registered
- Missing region (no param, no env vars) throws `InvalidOperationException` at registration time
- `AWS_REGION` env var is used as fallback when no explicit region param is given
- Explicit region param takes precedence over `AWS_REGION` env var

A real MSK cluster is required to validate the end-to-end SigV4 token lifecycle — manual test plan is at `tests/manual/68-kafka-msk-iam/test-plan.md`.

Closes #682


nightBaker (commented):
## Code Review — feat(streaming-kafka): AWS MSK IAM authentication (#682)

### 🟡 Important

1. **Multi-line inline comments in method bodies** — `MskIamTokenRefresher.cs` has two 2-line comment blocks; `MskIamSiloBuilderExtensions.cs` has one 2-line comment block. CLAUDE.md: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max." Each of these explains a genuine non-obvious constraint (async void fire-and-forget, dispose-race swallow, fail-fast ordering) and does survive the "would confuse a reader if removed" test — but CLAUDE.md caps them at one line. Condense each to a single line in a follow-up.

2. **XML docstrings on test helpers** — `MskIamSiloBuilderExtensionsTests` class and `FakeSiloBuilder` both carry multi-paragraph XML summaries. The production XML doc on `AddKafkaStreamingWithMskIam` follows the existing `KafkaProductionPresetExtensions` pattern (which has 5 such blocks), so it's consistent. The test-class docstrings aren't necessary for IntelliSense and should be trimmed or dropped.

3. **Env-var mutation without parallel isolation** — Tests 2 (`Add_without_region_or_env_throws`) and 3 (`Add_falls_back_to_AWS_REGION_env`) set `AWS_REGION`/`AWS_DEFAULT_REGION` with `finally`-restore. MSTest runs sequentially by default so it's fine today, but adding `[DoNotParallelize]` explicitly on those two methods would future-proof this.

### 🟢 Minor

1. **Unused `using System.Text;`** in `MskIamSiloBuilderExtensionsTests.cs` — remove it.

2. **`_ = Amazon.RegionEndpoint.GetBySystemName(effectiveRegion)` doesn't validate non-empty bad regions** — the inline comment says "Validate at extension-call time" but AWSSDK.Core 4.x returns `RegionEndpoint{DisplayName="Unknown"}` for any non-empty unrecognized string without throwing. Only the explicit `IsNullOrEmpty` guard above it actually validates. The comment should say "guard against empty" rather than implying all invalid regions are caught here.

3. **`new AWSMSKAuthTokenGenerator()` per refresh** — allocates a new instance on every OAUTHBEARER token refresh cycle. The class is stateless, so this is safe and correctness is unaffected; a static singleton would be a minor memory/GC win if token refresh becomes hot.

---

No blocking issues. All 4 unit tests pass; docs updated across contributor and website surfaces; EventId range correctly allocated at 11200–11299; async-void fire-and-forget pattern is correct for Confluent.Kafka's callback; double try/catch for the dispose-race is the right defensive move. OK to promote.

PR #739, opened 2026-06-18
feat(streaming-kafka): Schema Registry dual-encoding mechanism (#700)

nightBaker:
## Summary

Closes #700

- Adds `IExternalEventEncoder` interface to `Fleans.Streaming.Kafka` — single `Task<byte[]?> EncodeAsync(object, CancellationToken)` extension point for Schema Registry wire format
- Adds two-topic fanout to `KafkaQueueAdapter`: primary `fleans-N` (Orleans codec) + optional `fleans-N-events` (SR codec) when encoder is registered
- Adds `Fleans.Streaming.Kafka.SchemaRegistry` project — `KafkaSchemaRegistryOptions` (URL, Basic auth, TLS/mTLS) + `AddKafkaSchemaRegistry` extension that registers `ISchemaRegistryClient`
- Fanout is best-effort: EventId 11107 (encode fail) and 11108 (produce fail) log as Warning and are swallowed — primary Orleans delivery is unaffected

## Key decisions

- `IExternalEventEncoder` is **non-generic** to avoid per-event-type DI registrations; implementations check the event type internally and return `null` to skip
- Encoder wired via `services.GetService<IExternalEventEncoder>()` — null → no fanout, zero overhead default
- Constructor split: public ctor chains to `internal` via `BuildProducer`; internal ctor takes injectable `IProducer<byte[], byte[]>` + encoder for testability
- Bearer token auth removed — `SchemaRegistryConfig` in `Confluent.SchemaRegistry` does not expose this field; follow-up work can implement via a custom `ISchemaRegistryClient` wrapper if needed
- Avro (#685B) and Protobuf (#685C) framing are out of scope — each will install one `IExternalEventEncoder`

## Test plan

- `KafkaAdapterTests` (3 tests): no-encoder produces once, encoder-returns-null produces once, encoder-returns-bytes produces twice with second call targeting `-events` topic
- `SchemaRegistryTests` (2 tests): defaults validation, config binding
- Manual plan: `tests/manual/19-schema-registry-mechanism/test-plan.md` (requires Kafka + SR)

### DIFF
--- a/docs/conventions/observability-eventids.md
+++ b/docs/conventions/observability-eventids.md
@@ -22,7 +22,7 @@ propagation), see [`docs/plans/2026-02-08-structured-workflow-logging.md`](../pl
 | 9000–9099 | `BpmnConverter` | |
 | 10000–10099 | `TimerCallbackGrain` | |
 | 11000–11099 | `KafkaProductionPresetExtensions` | 11000 preset-applied INFO |
-| 11100–11199 | `KafkaQueueAdapter`, `KafkaQueueAdapterReceiver`, `KafkaQueueAdapterFactory` | 11100 SSL-no-paths OS-trust-store WARNING (shared EventId across all three sibling classes); 11101 acks-not-all WARNING; 11102 single-broker RF fallback INFO; 11103 RF>brokers fallback WARNING; 11104 GetMetadata error ERROR; 11105 topics created INFO; 11106 topics already existed INFO |
+| 11100–11199 | `KafkaQueueAdapter`, `KafkaQueueAdapterReceiver`, `KafkaQueueAdapterFactory` | 11100 SSL-no-paths OS-trust-store WARNING (shared EventId across all three sibling classes); 11101 acks-not-all WARNING; 11102 single-broker RF fallback INFO; 11103 RF>brokers fallback WARNING; 11104 GetMetadata error ERROR; 11105 topics created INFO; 11106 topics already existed INFO; 11107 fanout-encode-failed WARNING; 11108 fanout-produce-failed WARNING |
 | 11200–11299 | `MskIamTokenRefresher` | 11200 token-refresh-failed ERROR, 11201 set-token-failure-threw WARNING |
 
 ## Adding a new range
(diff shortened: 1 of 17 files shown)


nightBaker (commented):
-

nightBaker (commented):
## Code Review — feat(streaming-kafka): Schema Registry dual-encoding mechanism (#700)

### CLAUDE.md compliance ✅
- `[LoggerMessage]` source generators used exclusively; EventIds 11107/11108 correctly allocated in `observability-eventids.md`
- `private partial void` on `partial class KafkaQueueAdapter` — correct pattern
- Documentation updated across all four required surfaces (repo docs, website, EventId catalog, manual test plan)
- Tests follow MSTest + AAA pattern; 3 adapter tests + 2 options tests

### Code quality ✅
- Constructor split (public → `BuildProducer` → internal) is clean; testability overload correctly `internal` with `InternalsVisibleTo` in both csproj and `DynamicProxyGenAssembly2`
- Fanout placed after the primary produce's catch-rethrow block — correct: failed primary delivery silently skips fanout entirely
- `[LoggerMessage]` parameter name–to–message-placeholder matching is correct (`{EventType}`/`{Topic}` match `eventType`/`topic` params by name, not position)
- `IExternalEventEncoder` design rationale comment ("non-generic by design") is an appropriate WHY comment per CLAUDE.md
- `AllExpectedEventTopics` correctly mirrors `AllExpectedTopics` and `EventsTopicForQueue` correctly delegates to `TopicForQueue`
- `EnsureTopicsAsync` union correctly gates on `_encoder is not null` — no spurious `-events` topics when no encoder registered

### 🟡 Important
1. **Folder name mismatch**: `tests/manual/19-schema-registry-mechanism/` should be `69-schema-registry-mechanism/` per CLAUDE.md's "Create `NN-feature-name/` (next sequential number)" rule — `68-kafka-msk-iam` is the predecessor, so next is `69-`. The prefix `19` is already used by `19-event-subprocess-error/`. The README entry (line 756) references the wrong folder path. Not a test breakage but breaks the convention and creates confusion. Suggest renaming in a follow-up (or squash-fix before merge).

### 🟢 Minor
1. `InternalsVisibleTo DynamicProxyGenAssembly2` in `Fleans.Streaming.Kafka.SchemaRegistry.csproj` is unnecessary — `SchemaRegistry.Tests` has no NSubstitute dependency. Harmless.
2. XML `<summary>` block in `KafkaSchemaRegistrySiloBuilderExtensions.cs` is 3 comment lines; CLAUDE.md says one short line max. Consider condensing or removing.
3. `IOptionsMonitor<T>.CurrentValue` is the hot-reload API; `IOptions<T>` is more semantically correct for a one-shot singleton factory. Minor allocations difference, not a correctness issue.
4. No `Url`-empty guard before `CachedSchemaRegistryClient` construction — would produce an unhelpful exception at first SR call. Consider a startup validation (optional, consistent with how other options are validated).

